PR #25, opened 2016-11-30
Graph editor

Sheyne:
Change GraphView interfaces architecture to put interfaces each component expects with the component that requests them rather than as standalone entities. 

(Graph|Edge|Node)View interfaces are now called Drawable(Graph|Edge|Node) to reflect the fact that these interfaces are only the way that the graph editor sees them. Additionally they are all defined in the graph-editor.compontent file. graph-editor.compontent still pulls in the implementation of the Graph class, but this is temporary until we get everything better integrated.


scripts/main.component.ts
```diff
@@ -20,18 +21,36 @@ import { REPLComponent, REPLDelegate } from "./repl.component"
...
+  }
+  clearSelected() {
+    this.selectedElements.clear();
+  }
+
+
   ngOnInit(): void {
     this.menu.setDelegate(this);
     this.repl.setDelegate(this);
+    this.graphEditor.delegate = this;
```
RighteousRaichu: I did setDelegate()/removeDelegate() or something similar, as sometimes a class can have more than one delegate. (like I did with Menu service, but also that's more of an event. So maybe rename the menu as an event subscription and for delegates always assign to a delegate property?) What do you think, @Sheyne?

Sheyne: I vote either the way I did it, or `addDelegate`/`removeDelegate`. `set` implies theres only one

Sheyne: Honestly though, lets have delegates be a property, and anything else be called eventListeners

scripts/main.component.ts
```diff
@@ -20,18 +21,36 @@ import { REPLComponent, REPLDelegate } from "./repl.component"
   templateUrl: "../html/main.component.html",
   providers: [MenuService, PluginService]
 })
-export class MainComponent implements OnInit, MenuDelegate, REPLDelegate {
+export class MainComponent implements OnInit, MenuDelegate, REPLDelegate, EditorDelegate {
   constructor(private menu: MenuService, private pluginService: PluginService) {}
+  
+  selectedElements : Set<DrawableGraph | DrawableNode | DrawableEdge> =
+    new Set<DrawableGraph | DrawableNode | DrawableEdge>();
+  selectElement(element : DrawableGraph | DrawableNode | DrawableEdge) {
```
RighteousRaichu: I still think we should make the editor do the book keeping, and only inform the delegate of changes. For instance, what if there's no delegate set? Or if the delegate does something weird. Then the graph editor wouldn't work.

Sheyne: That could be worth doing. @ilargierdi had suggested making the main-component the source of truth, but we could switch it around. What do you thing @ilargierdi ?

RighteousRaichu: My opinion is that the main component AKA document controller mostly just wires everything up together. I want to avoid doing serious work in it. The document controller WILL end up being massive, and I'd prefer not to clutter it any further. Otherwise we're going to end up with a god component that all other's depend on (indirectly in this case, but still.)

Sheyne: I was going to put this somewhere else, but still owned by the main.controller. Do you think it should be owned by the editor? I think you might be right.

ilargierdi: The editor component originally did the bookkeeping for selected items. I'm not sure that I like the idea of having that logic in the component itself, but I can't think of a better place to put it. We can put that stuff back into the component and then have the component expose an event handler field `selectionchanged` or something that the outside world can bind to. I had a note about it in one of the comments.

RighteousRaichu: I propose we do exactly that. Put it back in the editor component and just expose a selectionChanged event.

PR #26, opened 2016-12-02
Added edge labels and fixed node labels

ilargierdi:
Edge labels are now drawn on the canvas, and node labels are now set correctly when a new node is created.


ilargierdi:
These changes should now be present in the saving branch.

PR #50, opened 2017-01-11
Autoformatting

Sheyne:
I'd like everyone to sign off that they agree this is a good idea before we merge it. I like always auto-formatting the project, so we don't have commits where one person changes someone else formatting. 

I'm not sure adding auto format to the build script is the right place to do it, because currently `npm run build` doesn't effect the ./app dir. 

I want it to be really hard to commit changes without auto formatting, so maybe a commit hook? 

The one thing I don't like about how this thing formats for functions with lots of options. I agree that each option should be on its own line, but they should be indented up to the opening paren. 

Thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,11 +10,11 @@
   "main": "index.js",
   "scripts": {
     "clean": "rm -r ./node_modules 2> /dev/null || echo > /dev/null && rm -r ./build 2> /dev/null || echo > /dev/null && rm -r ./dist 2> /dev/null || echo > /dev/null",
-    "build": "mkdir build > /dev/null || echo > /dev/null && npm install && webpack --progress --profile --colors --display-error-details --display-cached && cp package.json build",
+    "build": "tsfmt -r && mkdir build > /dev/null || echo > /dev/null && npm install && webpack --progress --profile --colors --display-error-details --display-cached && cp package.json build",
     "start": "electron build",
     "start:build": "npm run build && npm run start",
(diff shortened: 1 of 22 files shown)


ilargierdi:
I am okay with this. In regards to the indentation on function parameters, I prefer the way the autoformatter handles it better than aligning the list to the opening paren.

RighteousRaichu:
I don't particularly care either way, but I'm not against it. I agree with CJ on moving it after `npm install`

package.json
```diff
@@ -13,17 +13,18 @@
...
+    "prebuild": "mkdir build > /dev/null || echo > /dev/null && mkdir dist > /dev/null || echo > /dev/null && npm install",
+    "build": "npm run prebuild && npm run format && npm run webpack",
     "build:watch": "npm run prebuild && npm run webpack -- --watch",
 
     "start": "electron build",
     "start:build": "npm run build && npm run start",
 
     "electron-packager": "electron-packager build sinap --out=dist --no-prune --arch=all --overwrite",
     "prepackage": "npm run clean && npm run build && cp package.json build",
-    "package":"npm run prepackage && npm run electron-packager -- --platform=all",
```
Sheyne: I have my sublime set to auto-format JSON on save. This is not part of the tsfmt, but it keeps touching this file when I edit it.

RighteousRaichu: I'll be more careful when I edit it to be consistent. But you should really start using the one true code editor AKA VS Code ;)

Sheyne: I thought we wanted to use different editors to ensure that other devs would run into stuff that was VS Code specific. If it bugs you I can turn off my JSON plugin.

PR #52, opened 2017-01-14
Graph editor refactor

ilargierdi:
Moved a lot of stuff around and changed the drawing logic.


Sheyne:
Does this expose quadratic lines? Or are they still in development?

slaymaker1907:
Why is NODE_FONT_FAMILY a separate variable instead of being in the display property defaults?

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -0,0 +1,398 @@
...
+    return result;
+}
+
+/**
+ * isEdgeOverlapped  
+ *   Checks if two edges are overlapping.  
+ *   It is assumed that the edge being checked does not have the same source
+ *   and destination nodes.
+ */
+export function isEdgeOverlapped(e: DrawableEdge, nodeEdges: Map<DrawableNode, Set<DrawableEdge>>): boolean {
```
slaymaker1907: We should probably not allow users to try and connect two nodes with the same source and destination either by error message or by over writing the old edge. Ordinarily, I would say we should allow them to make such a thing, but this would violate the mathematical properties of graphs and does not seem to be a common use case.

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -0,0 +1,398 @@
...
+    clone.lineWidth = e.lineWidth;
+    clone.label = e.label;
+    return clone;
+}
+
+/**
+ * cloneNode  
+ *   Creates a cloned node.
+ */
+export function cloneNode(n: DrawableNode): DrawableNode {
```
slaymaker1907: In my opinion, this should be replaced with a generic cloning function if possible or be a method of the DrawableNode class. i.e., it should only depend on the contents of DrawableNode if it is a method else it should generically do a deep copy.

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -0,0 +1,398 @@
...
+     * canCreateEdge  
+     *   Checks if a drawable edge can be created with the given source and
+     *   destination nodes.
+     * 
+     *   If `like` is specified, a drawable edge with a matching type of `like` is
+     *   is checked against the given source and destination nodes.
+     * 
+     *   This method must be called before calling `createEdge`.
+     */
+    canCreateEdge(
```
slaymaker1907: We should have this method available, but createEdge should probably also check this and throw an exception if the edge can not be created unless that check is too expensive for some reason.

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -0,0 +1,398 @@
...
+        src: DrawableNode,
+        dst: DrawableNode,
+        like?: DrawableEdge
+    ): DrawableEdge;
+
+    /**
+     * replaceEdge  
+     *   This method has been replaced by `createEdge` and is no longer needed.
+     */
+    replaceEdge(original: DrawableEdge, replacement: DrawableEdge): void;
```
slaymaker1907: This should be removed since this is a refactor or the comment should be updated.

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -0,0 +1,398 @@
...
+ * DrawableEdge  
+ *   Interface that exposes drawable edge properties.
+ */
+export interface DrawableEdge {
+
+    /**
+     * source  
+     *   The source node of the edge.
+     */
+    source: DrawableNode | null;
```
slaymaker1907: Should clarify in comments under what conditions source or destination can be null.

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -0,0 +1,398 @@
...
+     * color  
+     *   The color of the edge. This can be any valid `CSS` color string.
+     */
+    color: string;
+
+    /**
+     * lineStyle  
+     *   The line style of the edge. This can be `solid`, `dotted`, or `dashed`.
+     */
+    lineStyle: string;
```
slaymaker1907: This should be an enum.

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -0,0 +1,398 @@
...
+     * label  
+     *   The text label to be displayed in the node.
+     */
+    label: string;
+
+    /**
+     * shape  
+     *   The shape of the node.
+     */
+    shape: string; // TODO: for now this only supports circles and squares.
```
slaymaker1907: This should be an enum.

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -0,0 +1,398 @@
...
+     * color  
+     *   The background color of the node.  Can be any valid `CSS` color string.
+     */
+    color: string;
+
+    /**
+     * borderColor  
+     *   The border color of the node.  Can be any valid `CSS` color string.
+     */
+    borderColor: string;
```
slaymaker1907: Colors should probably be a number. This can then be converted to and from hex for io.

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -0,0 +1,398 @@
...
+
+    // TODO:
+    // more display properties.
+}
+
+/**
+ * DrawableNode  
+ *   Interface that exposes drawable node properties.
+ */
+export interface DrawableNode {
```
slaymaker1907: There is a lot of duplicate stuff in DrawableNode and DrawableEdge. It may be a good idea to encapsulate some of these as a separate class and use object composition (or inheritance if that makes more sense).

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -0,0 +1,398 @@
...
+}
+
+// Type aliases ////////////////////////////////////////////////////////////////
+
+
+/**
+ * Drawable  
+ *   Type alias for the the union type of `DrawableEdge` and `DrawableNode`.
+ */
+export type Drawable = DrawableEdge | DrawableNode;
```
slaymaker1907: Since you are using them as the same thing anyway for some functions, it might be worth making Drawable a superclass of DrawableEdge and DrawableNode.

app/components/graph-editor/make-fn-node.ts
```diff
@@ -0,0 +1,222 @@
...
+
+import * as CONST from "./constants";
+import * as canvas from "./canvas";
+import { DrawableNode } from "./drawable-interfaces";
+
+
+/**
+ * makeFnNode  
+ */
+export function makeFnNode(
```
slaymaker1907: This seems like a fairly complex bit of code and should maybe have some more comments.

app/components/graph-editor/math.ts
```diff
@@ -0,0 +1,29 @@
...
+// Date created: January 9, 2016
+//
+// THIS FILE IS INTENDED TO BE IMPORTED ONLY INTO graph-editor.component.ts
+
+
+/**
+ * dot  
+ *   Calculates the dot product of two vectors.
+ */
+export function dot(a: number[], b: number[]): number {
```
slaymaker1907: This seems like it could be useful for the ML plugin if we decide to have an option for running as JS. Also, this seems like potentially a good place to bring in a library for linear algebra stuff if only for performance. Something like vectorius.

app/components/main/main.component.ts
```diff
@@ -79,20 +79,17 @@ export class MainComponent implements OnInit, MenuEventListener, REPLDelegate, T
...
-    onGraphChanged = () => { // arrow syntax to bind correct "this"
-        if (this.graph) {
+    onContextChanged = () => { // arrow syntax to bind correct "this"
+        if (this.context) {
             if (this.graphEditor) {
                 this.graphEditor.redraw();
             }
             if (this.pluginService) {
-                if (this.graph.pluginManager.kind == "machine-learning.sinap.graph-kind") {
+                if (this.context.graph.pluginManager.kind == "machine-learning.sinap.graph-kind") {
```
slaymaker1907: This seems like a magic number (in this case a string).

app/components/main/main.component.ts
```diff
@@ -218,7 +219,12 @@ export class MainComponent implements OnInit, MenuEventListener, REPLDelegate, T
                 break;
             }
         } else {
-            newSelectedEntity = this.graph;
+            if (this.context) {
+                newSelectedEntity = this.context.graph;
+            } else {
+                throw "How did graph selection change, there's no context? ";
```
slaymaker1907: This should be an assert check since this is basically a buggy code check.

app/components/graph-editor/drawable-interfaces.ts
ilargierdi: I'm not quite sure what you're arguing here. Edges can have the same source and destination node in general; it just turns out that the node is self-referencing. If the graph language does not support self-referencing nodes, that behavior should be reflected in the implementation of the `DrawableGraph.canCreateEdge` method. As an aside, what you are proposing is the concern of the implementation of `DrawableGraph.canCreateEdge` and not the graph editor.

app/components/graph-editor/drawable-interfaces.ts
ilargierdi: I agree that it should be part of the `DrawableNode` class, and same for the `cloneEdge` function. As for deep copying, I don't think that's necessary; at least, not for how it's being used in the graph editor.

app/components/graph-editor/drawable-interfaces.ts
ilargierdi: I mentioned this to Sheyne last semester. He wanted it this way.

app/components/graph-editor/drawable-interfaces.ts
ilargierdi: The code refactor was primarily for the graph editor component, not the drawable interfaces. I'll remove this on the next push.

app/components/graph-editor/drawable-interfaces.ts
ilargierdi: I am aware.

app/components/graph-editor/drawable-interfaces.ts
ilargierdi: I am also aware of this as well. I was thinking of doing something along the lines of union string types; e.g. `"circle" | "square" | "image"`, I just haven't gotten around to it.

app/components/graph-editor/drawable-interfaces.ts
ilargierdi: It's a string because it's easier to set the canvas rendering context `.strokeStyle` this way.

app/components/graph-editor/drawable-interfaces.ts
ilargierdi: This should be a separate git issue.

app/components/graph-editor/make-fn-node.ts
ilargierdi: `graph-editor.component.ts` line 49.

app/components/graph-editor/math.ts
ilargierdi: A linear algebra library would be nice.

app/components/main/main.component.ts
ilargierdi: I wasn't the one who did this. Ask Sheyne about it.

app/components/main/main.component.ts
ilargierdi: Same as comment above.

app/components/graph-editor/drawable-interfaces.ts
ilargierdi: I may change this to be a definite `DrawableNode`, since it'll be easier to not draw hidden graph elements with the code refactor in place. Originally, the `| null` was so that I can have one end of the edge follow the cursor as it is being dragged in the case of moving an edge from one node to another.

app/components/graph-editor/drawable-interfaces.ts
ilargierdi: Same comment as above.

ilargierdi:
@Sheyne 
> Does this expose quadratic lines? Or are they still in development?
Quadratic lines are currently only used for overlapped edges.

@slaymaker1907 
> Why is NODE_FONT_FAMILY a separate variable instead of being in the display property defaults?
That is a good question.

> isItThat should maybe handle the case if "that" is a falsy type like null. i.e., isItThat should probably return true if it == that == null.
> Also, I think isItThat will fail for nested objects.
`isItThat` is not exposed outside of `drawable-interfaces.ts`; it is a helper function used only to determine if `it` has the same properties and methods along with their associated types as `that`. I am aware of the problems you mention, but the function is meant to only look at the presence of properties and methods as opposed to their contents.

ilargierdi:
Something I should probably mention:

None of the extra files are intended to be imported outside of `graph-editor.component.ts`.

app/components/main/main.component.ts
RighteousRaichu: I think this is just temporary until we have plugins working.

app/components/graph-editor/drawable-interfaces.ts
RighteousRaichu: I think this should entirely be up to canCreateEdge, also we need to figure out how we want to expose this to the current plugin. (For this pull request it's fine.)

PR #53, opened 2017-01-16
Plugin API Update and File IO Service

slaymaker1907:
I changed the plugin API a bit to be simpler and separated file IO into a separate service (so that we can potentially move it to the web if desired for a demo).

Plugin loading is not yet dynamic or sandboxed, but that will be the next step and was the reason that file IO was redone since that will likely be necessary. I'm leaving in the sandbox folder for now since it will serve as a good example for how to safely run plugins.

One other item of note is that while PluginManager is currently an interface, I think it would make more sense to have that be loaded from a config file for simplicity as well as tightening up plugin security (i.e. we should only run JS, even if sandboxed, when absolutely necessary).


Sheyne:
I'll review the rest of the PR in a bit. With regards to PluginManager: PM is an interface that needs to change, but it's basic functionality will stick around. The objects implementing it will be coded by us, and will be loaded from the plugin's configuration. That is, PM doesn't correspond to external code.

RighteousRaichu:
The File IO Service is beautiful. Eventually we might want to wrap up files in a more substantial way. (When we start building out the files panel. For example we'd want a notion of which files are open and which have been modified since being saved and whatnot.) That might be it's own service (which would use the File IO Service.) It'd be especially nice because we could theoretically swap the File IO service with say, a Google Drive service.

Issue #19, opened 2016-11-23
Add Ability for Nodes to Self Reference


slaymaker1907:
Right now you can only connect two different nodes.

slaymaker1907:
This fulfills part of issue #2

ilargierdi:
Right now, edges do not make a distinction between source and destination nodes. If both source and destination nodes are equal, the edge should be drawn as a loop. I'm leaving this comment for future reference when I get around to implementing the change. In general, though, the `DrawableGraph.canCreateEdge` interface method should determine whether or not an edge can loop back into its source node.

PR #55, opened 2017-01-20
Issue19

ilargierdi:
Only review if you're motivated to. Refer to the commit message from b9f0ee4 for a summary of changes.

This branch mainly addresses #19 .

### DIFF
--- a/app/components/graph-editor/canvas.ts
+++ b/app/components/graph-editor/canvas.ts
@@ -59,7 +59,7 @@ export function drawLine(
 
 /**
  * drawQuadraticLine  
- *   Draws a quadratic line between two points.
+ *   Draws a quadratic bezier line between two points.
  */
 export function drawQuadraticLine(
     g: CanvasRenderingContext2D,
@@ -76,7 +76,11 @@ export function drawQuadraticLine(
     g.stroke();
 }
 
-export function drawBezierLine(
+/**
+ * drawCubicLine  
+ *   Draws a cubic bezier line between two points.
+ */
+export function drawCubicLine(
     g: CanvasRenderingContext2D,
     src: number[],
     dst: number[],
@@ -95,13 +99,18 @@ export function drawBezierLine(
 
 /**
  * drawArrow  
- *   Draws an arrow towards the destination node.
+ *   Draws an arrow towards the destination point.
+ * 
+ *   The arrow is drawn by computing the unit vector from the given source and
+ *   destination points and rotating, scaling, and translating the unit vector
+ *   before drawing the left and right sides of the arrow.
  */
 export function drawArrow(
     g: CanvasRenderingContext2D,
     src: number[],
     dst: number[]
 ): void {
+    // Get the unit vector from the source point to the destination point.
     let v = [
         dst[0] - src[0],
         dst[1] - src[1]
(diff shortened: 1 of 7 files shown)


app/components/graph-editor/canvas.ts
```diff
@@ -77,14 +77,40 @@ export function drawQuadraticLine(
 }
 
 /**
+ * drawCubicLine  
+ *   Draws a cubic bezier line between two points.
+ */
+export function drawCubicLine(
```
slaymaker1907: It is not clear what these parameters do.

app/components/graph-editor/canvas.ts
```diff
@@ -269,14 +295,19 @@ export function setLineStyle(
...
-    if (!dotSize)
-        dotSize = g.lineWidth;
-    if (value == "dashed")
-        g.setLineDash([3 * dotSize, 6 * dotSize]);
-    else if (value == "dotted")
-        g.setLineDash([dotSize, 2 * dotSize]);
-    else
-        g.setLineDash([1, 0]);
+    dotSize = (dotSize ? dotSize : g.lineWidth);
+    switch (value) {
```
slaymaker1907: Maybe consider making value a function that draws the line given dotSize and g.

app/components/graph-editor/canvas.ts
```diff
@@ -379,95 +410,153 @@ export function getStraightEdgePoints(
...
+export function getLoopEdgePoints(
+    e: DrawableEdge,
+    src: DrawableNode,
+    srcDim: any
+): number[][] {
+    let u = [MathEx.SIN_22_5, -MathEx.COS_22_5];
+    let v = [-MathEx.SIN_22_5, -MathEx.COS_22_5];
+    let pt0 = getEdgePtShift(u, src, srcDim);
+    let pt1 = getEdgePtShift(v, src, srcDim);
+    let pt2 = [
```
slaymaker1907: Extract this out to a function to avoid duplicate code between pt2 and pt3.

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -396,3 +286,127 @@ export class DefaultEdge implements DrawableEdge {
...
+    let result = true;
+    for (let p in that) {
+        result =
+            result &&
+            (p in it) &&
+            (typeof that[p] === typeof it[p]);
+        if (!result)
+            return false;
+    }
+    return result;
```
slaymaker1907: Just return true rather than result and just do a check in the for loop. result is an unnecessary variable.

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -396,3 +286,127 @@ export class DefaultEdge implements DrawableEdge {
...
+}
+
+/**
+ * isEdgeOverlapped  
+ *   Checks if two edges are overlapping.  
+ *   It is assumed that the edge being checked does not have the same source
+ *   and destination nodes.
+ */
+export function isEdgeOverlapped(e: DrawableEdge, nodeEdges: Map<DrawableNode, Set<DrawableEdge>>): boolean {
```
slaymaker1907: Combine the two sets of edges into one iterable since the two loops are otherwise the same. Another option would be to extract the predicate of the loop into a function.

app/components/graph-editor/canvas.ts
ilargierdi: This comment seems kind of petty. Please refer to the third bullet point of the commit message for b9f0ee4.

slaymaker1907: Sounds good, I just didn't really know what was going on.

app/components/graph-editor/canvas.ts
ilargierdi: `value` is supposed to come straight off of a graph element. I'm not exactly sure what the advantage would be of making value a function that returns the appropriate array for `g.setLineDash`. The function would still have a switch or if/else statement in it.

app/components/graph-editor/canvas.ts
ilargierdi: The code you are referring to is only used in `getLoopEdgePoints`. Putting it into a function would probably save three lines, but expand the file by more to take into account the header comment for the function. It wouldn't make sense to use the new function outside of `getLoopEdgePoints`, because it is specific to the control points `pt2` and `pt3`. I could make the new function more general, but that would require adding more general functions in order to do what is essentially a vector scale and translation. I don't think expanding the stack for such a calculation would be worth it. If your main concern is readability, note that there is very minimal inline documentation. The full implementation of graph editor is not complete, and I intend to fully document once all of the other minor issues are addressed.

PR #58, opened 2017-01-22
Graph.ts refactor

Sheyne:


app/components/main/main.component.ts
```diff
@@ -83,11 +85,11 @@ export class MainComponent implements OnInit, MenuEventListener, REPLDelegate, T
                 this.graphEditor.redraw();
             }
             if (this.pluginService) {
-                if (this.context.graph.pluginManager.kind == "machine-learning.sinap.graph-kind") {
+                if (this.context.graph.graph.plugin.kind == "machine-learning.sinap.graph-kind") {
```
ilargierdi: `context.graph.graph` looks kind of funny.

app/components/main/main.component.ts
```diff
@@ -217,6 +221,16 @@ export class MainComponent implements OnInit, MenuEventListener, REPLDelegate, T
         }
     }
 
+    propertyChanged(event: [PropertiedEntity, keyof PropertiedEntityLists, string, string[]]) {
+        // THIS IS SUPER DIRTY AND CJ SHOULD REALLY HOOK THE CHANGE DETECTOR
+        // TODO: KILL THIS WITH FIRE
```
ilargierdi: Alright, already, jeez.

app/models/drawable.ts
```diff
@@ -0,0 +1,51 @@
+// File: graph.ts
```
ilargierdi: Not really important, but this should probably be updated. Also adding a `Contributor` field would be nice, but not necessary.

app/components/main/main.component.ts
Sheyne: I'm sorry.

app/components/properties-panel/properties-panel.component.ts
```diff
@@ -3,32 +3,64 @@
...
+
+    @Output() propertyChanged = new EventEmitter();
+
+    // TODO: 
+    // discuss with Daniel
+    // I want to wrap up this functionality into bind.directive
+    // so that we can just do:
+    // <input [snpBind]="selectedEntity" [group]="..." [key]="..." [keyPath]="...">
+    // I've mostly got it, but it needs to somehow link in the forms module
+    // I'll talk more in person.
```
ilargierdi: Just bringing this to @RighteousRaichu's attention.

RighteousRaichu: Yeah @Sheyne this is something we might want to sit down together and figure out. This area of angular changed considerable from angular 1.x so I'm as lost as you are as to the best way to accomplish this in angular 2.x. 
(Here's a resource explaining some of the changes: https://blog.thoughtram.io/angular/2016/10/13/two-way-data-binding-in-angular-2.html)

PR #61, opened 2017-01-25
Type updates

Sheyne:
Types is now rewritten from the ground up.

Classes now support (multiple) inheritance. 
   - multiple inheritance is a good thing because what I am calling classes are really just interfaces

Types can now be parsed from literals

```typescript
Type.parseType("class { x: String }")
Type.parseType("List<Number>")
Type.parseScope(`
DFANode = class Node {
    AcceptState: Boolean
}
`)
```


package.json
```diff
@@ -15,7 +15,7 @@
      "peg": "pegjs --allowed-start-rules Definitions,Type --dependency \"./type-structures\" app/types/type-parser.pegjs && mv app/types/type-parser.js app/types/type-parser-inner.js && sed -i \"\" 's/^var \\.\\/type-structures/var types/' app/types/type-parser-inner.js",
```
Sheyne: I realize that this SED script is ridiculous, not sure there's a better way for now. 

This currently copies the '.js' file directly into the "app/" folder. @RighteousRaichu do you know how to fix this?

app/components/properties-panel/properties-panel.component.html
```diff
@@ -33,23 +26,32 @@
...
+                                class="text-input tuple-input"/>
+                            </template>
+
+                            <template [ngIf]="pair[1].subtype(Type.Node)">
+                                <label  class="node-label">
+                                    {{selectedEntity[group[1]].get(pair[0]).drawableProperties.get('Label')}}
+                                </label>
+                            </template>
+
+                            <!-- add neither case-->
```
Sheyne: Maybe import the elif package?

app/types/type-parser.pegjs
```diff
@@ -0,0 +1,66 @@
+// This represents a scope
+// currently it doesn't make any sense to have a nested scope
+
+Definitions = _ "{" _ dfns:Definition* _ "}" _ { return new types.TypeScope(new Map(dfns)); }
```
slaymaker1907: If we are making our own language,

```
return new types.TypeScope(new Map(dfns));
```
Seems much too verbose. Maintaining similarity to JavaScript is good, but we should take advantage of having our own parser where possible. For instance, at the very least we should remove the both "new"s from the spec.

slaymaker1907: With this comment and any others I make, please let me know if I'm just misunderstanding this parser generator.

app/types/type-structures.ts
```diff
@@ -0,0 +1,317 @@
...
+            this.definitions.set(key, type.feed(types));
+        }
+        return this;
+    }
+}
+
+export class TypeVariable implements Type {
+    kind = "Variable";
+
+    constructor(public type?: Type, public lookupName?: string, public matchName?: string) {
```
slaymaker1907: Why are all of these optional? It seems like these should be required fields intuitively. If partial initialization is needed, JSON may be more appropriate until fully initialized.

app/types/type-structures.ts
```diff
@@ -0,0 +1,317 @@
...
+            return true;
+        }
+        if (that.subtype(this.type)) {
+            this.type = that;
+            return true;
+        }
+        return false;
+    }
+
+    rsubtype(that: Type) {
```
slaymaker1907: Personally, I would have subtypeing return a new type so that you still have access to the old one if needed and it seems like it wouldn't be too much of a performance hit.

I also don't like combining read and write methods.

app/types/type-structures.ts
```diff
@@ -0,0 +1,317 @@
...
+            if (replacementType) {
+                return replacementType;
+            }
+        }
+        return this;
+    }
+    validate() { }
+
+    toString() {
+        return "(" + this.matchName + "?" + this.lookupName + ":" + this.type + ")";
```
slaymaker1907: Maybe change this out with an ES6 template string for clarity.

app/types/type-structures.ts
```diff
@@ -0,0 +1,317 @@
...
+        const myPromise = this.fields.get(key);
+        if (myPromise) {
+            result.push(myPromise);
+        }
+        return result;
+    }
+
+    subtype(that: Type): boolean {
+        if (that.rsubtype) {
+            return that.rsubtype(this);
```
slaymaker1907: Couldn't this potentially mutate that?

app/types/type-structures.ts
```diff
@@ -0,0 +1,317 @@
...
+
+    feed(types: Map<string, Type>) {
+        this.from = this.from.feed(types);
+        this.to = this.to.feed(types);
+        return this;
+    }
+    validate() { }
+}
+
+export class ThunkType implements Type {
```
slaymaker1907: What is a Thunk?

package.json
```diff
@@ -12,9 +12,10 @@
     "clean": "rm -r ./build 2> /dev/null || echo > /dev/null && rm -r ./dist 2> /dev/null || echo > /dev/null",
     "clean:node_modules": "npm run clean && rm -r ./node_modules 2> /dev/null || echo > /dev/null",
     "webpack": "webpack --progress --profile --colors --display-error-details --display-cached",
+     "peg": "pegjs --allowed-start-rules Definitions,Type --dependency \"./type-structures\" app/types/type-parser.pegjs && mv app/types/type-parser.js app/types/type-parser-inner.js && sed -i \"\" 's/^var \\.\\/type-structures/var types/' app/types/type-parser-inner.js",
```
slaymaker1907: This should have a windows version if possible.

package.json
RighteousRaichu: `"pegjs --allowed-start-rules Definitions,Type --dependency \"types:./type-structures\" app/types/type-parser.pegjs && mv app/types/type-parser.js app/types/type-parser-inner.js"`

Is this what you're asking for?

And what do you mean it copies it to the "app/" folder? It looks like the mv command just renames it?

Sheyne:
@RighteousRaichu will you read through the inline comments I added and the README.md and let me know whether you think that's sufficient? 

Also @slaymaker1907 I think that I've addressed most of your confusion about syntax in the readme.

app/types/type-structures.ts
Sheyne: Zero argument function, used to delay execution. Will probably be used for computed fields.

app/types/type-structures.ts
Sheyne: Yes, but only if its a type variable.

app/types/type-structures.ts
Sheyne: Subtype is not a create method, it (as wasn't documented when you commented but is now) tests whether `this` is a subtype of `that`. Mutation only happens if it is a type variable in type inference. `rsubtype` is the reverse, in principle: 

```TypeScript
this.subtype(that) === that.subtype(this)
```
Essentially I want things that define a `rsubtype` to have more say in the process.

app/types/type-structures.ts
Sheyne: TypeVariables are used in two contexts, type inference, and type referencing. Either can define a match name. All three optional can be defined, or none of them, things will still work.

app/types/type-parser.pegjs
Sheyne: If you read the app/types/README.md that I just committed, there are several examples of syntax. 

Link to file in the commit [here](https://github.com/2graphic/sinap-ide/blob/5f8d11e4786862c1b781126de619dca168137984/app/types/README.md). (This directly references the commit so people from the future be warned: it might of changed)

PR #66, opened 2017-01-25
Sandbox and dynamic loading

slaymaker1907:
This is still somewhat a work in progress but I wanted to get stuff merged in so that stuff doesn't diverge too far (specifically the plugin API has some cleaning up like removing the promise requirement for plugin files).

Note that plugins have been isolated a separate WebPack build. Also, the plugin does not currently import stuff from the main project since otherwise a bunch of unnecessary code is compiled into the generated js file due to WebPack due to recursive imports.


app/models/plugin.ts
```diff
@@ -63,48 +61,109 @@ export interface Program {
...
+export interface Edge {
+    Label: string;
+    Source: Node;
+    Destination: Node;
+    [propName: string]: any;
 }
 
 /**
  * This class is still in progress. Presumably, the InterpreterGraph will have different needs from the GUIGraph. TODO: Actually make this different.
  */
```
Sheyne: This is no longer a TODO

slaymaker1907: Updated documentation in the latest commit to sandbox.

webpack.config.js
```diff
@@ -139,4 +139,36 @@ module.exports = [
...
+            library: 'module'
+        },
+        resolve: {
+            extensions: ['.ts']
+        },
+
+        module: {
+            loaders: [
+                {
+                    test: /dfa-interpreter.ts$/,
```
RighteousRaichu: This can probably be all .ts files.
The way webpack works is it looks at your entry, then all of its requires, and so on... All those files are then sent through the loaders.
So specifying all .ts files here won't ruin anything. (probably.)

plugins/dfa-interpreter.ts
```diff
@@ -0,0 +1,86 @@
+declare var Interpreter: any, InterpreterGraph: any, Program: any, ProgramInput: any, ProgramOutput: any, RunningProgram: any;
+//import { Graph, Node, Edge } from "../app/models/plugin";
```
RighteousRaichu: This is the correct (hack) approach right now, right? Since we don't want the plugins to reference the main project at all. We need someone to declare these models in a way that both sinap and plugins can compile with them.

RighteousRaichu:
I don't necessarily think removing promises for the sake of removing promises is a good idea. IMO promises are one of the best ways in the Javascript world for dealing with Async code. And since the interpreters will be running in a different process then they are async. Overall I like the approach you've taken with the promises so far. Though if you have a cleaner solution in mind then it's probably fine, too.

PR #69, opened 2017-01-27
Refactor plugin loading

Sheyne:
This is it's own PR because it is the starting point for both everything I am about to do and Dyllon is about to do.


app/services/plugin.service.ts
```diff
@@ -142,54 +144,56 @@ export class PluginService {
...
+                .runInContext(context)
+                .then<Program>((program) => {
+                    return {
+                        run: (input: ProgramInput): Promise<ProgramOutput> => {
+                            context.sinap.__input = input;
+                            // Cast is necessary. 
+                            // I prefer forcing it to be explicit (and thus not any)
+                            return this.runInputCode.runInContext(context) as Promise<ProgramOutput>;
+                        },
+                        compilationMessages: [""]
```
slaymaker1907: We can actually get the compilationMessages since they are being produced by the plugin.

```
var script = this.sandboxService.compileScript('sinap.__program.then((program) => program.compilationMessages');
messagePromise = script.runInContext(context)
```
Not sure exactly how to work around the promise API here, though. Will probably be easier once the plugin API is updated to not require promises of plugin developers.

RighteousRaichu:
@Sheyne @slaymaker1907 
This branch seems to have some lag issues (and master doesn't), IDK if you're aware. Running a simple DFA takes a second or so, as well as creating new tabs (after running a DFA first.)
We can either open an issue and address the lag later or address it here. Then I'll merge this.

Sheyne:
This is really weird. The lag accumulates. If you reload the project, create a simple DFA, and run, the first 2 or 3 times you invoke the interpreter, everything is fast, by 4-5 it feels a little laggy, and it gets much worse after that. I'll go see if this occurs before my refactor.

Sheyne:
I think this is definitely a blocker for merging. Since I cant review, I've asked @RighteousRaichu to give it a failing review so we can't merge it until it's fixed.

app/services/plugin.service.ts
```diff
@@ -142,54 +144,56 @@ export class PluginService {
...
-        } else {
-            result = this.loadPlugin(kind, "./build/plugins/dfa-interpreter.js") // TODO: Put real file in here.
-                .then((cont) => {
-                    context = cont;
-                    this.interpreters.set(kind, context);
-                    return processContext(context);
+    public getInterpreter(graph: Core.Graph): Promise<Program> {
+        // TODO: consider: this could really just be a cast and 
+        // graph.plugin.context
+        return this.getPlugin(graph.plugin.kind).context.then((context) => {
```
Sheyne: Interesting reading: https://github.com/angular/zone.js/issues/130

slaymaker1907: I checked out the branch and can't seem to replicate this on Windows 10.

slaymaker1907: Also for reference, I'm running npm version 3.10.8 and node version 6.9.1.

slaymaker1907: Issue was that it did not build on Windows. I fixed the build issue and can now confirm the lag issue on Windows.

slaymaker1907: The issue appears to have been trying to cache plugins inside of a map. We should investigate why this is the case since it would be ideal to avoid file reading and compiling graphs as much as possible.

Issue #40, opened 2016-12-08
Implement zooming and panning in the graph editor component.


ilargierdi:
For input gestures, pinching should affect zoom and two-touch dragging should affect pan.

ilargierdi:
It turns out none of my development machines trigger [pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events), so I can't test for multi-touch gestures.

Sheyne:
Googling turns up: http://hammerjs.github.io

In some quick tests on my phone I was partially happy with its handling (except rotate which it really failed on). It allows emulation of touch events (via shift key) so maybe you can use that for dev?

Sheyne:
Also the comments here seem really useful. 

https://github.com/mapbox/mapbox-gl-js/issues/2618

ilargierdi:
I'm closing this issue and making a new one for two-touch input gestures.

PR #70, opened 2017-01-29
Zoom and Pan

ilargierdi:
This PR is mainly to push some changes from an earlier commit that may cause conflicts outside of the graph editor component. It mainly addresses #40.

For now, zoom is mapped to vertical scrolling and pan is mapped to the right mouse button until multi-touch gestures are implemented.


ilargierdi:
Yes, I should be able to. This brings up another point, though. When the user is trying to create a node, the destination node that the edge is hovering over should display some visual cue that the edge can successfully be created or not. I was thinking of just having the node have its color brightened for valid and darkened for invalid. In addition to this, I can also make it show the anchor point that the edge will snap to. Or, I can forgo the node highlighting and only show the anchor point for a valid edge drop but keep the darkening of the node for an invalid edge drop.

Sheyne:
The second one was more what I was thinking.

RighteousRaichu:
I like the second one, but maybe just show anchor and don't do any highlighting or darkening?

ilargierdi:
So show anchor for valid and hide anchor for invalid? I kind of like the idea of darkening or graying out the node for invalid, since it might be difficult to tell if the anchor point is present or not.

Sheyne:
What about an animation? Either:

* Flash the node blue on okay to add an edge and red on not okay
* Flash the anchor point 
    * and maybe make it grow from 0% size to 200% size down to 100% size) when you are making it appear
* maybe make the anchor point appear regardless, but greyed out if connection is not allowed.

RighteousRaichu:
I'm not sure on the animations. I think at a minimum just show/hide the anchors and I'd be happy. It doesn't hurt to try more though.

ilargierdi:
I'd have to think a little bit more about the animations. I'm not sure if it'd be worth the time and effort to implement. Maybe for 1.0 as a rank 3 feature, but only if I run out of things to do; although, I would rather be writing plugins or working on other areas of the UI. For now, I'll just have the anchor point either show or not depending on the result of `canCreateEdge`.

app/components/graph-editor/README.md
```diff
@@ -0,0 +1,82 @@
+# Notes
```
slaymaker1907: Really nice documentation. We should try and emulate these kinds of READMEs throughout the project! However, we should probably have a link from the main README to this README?

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -61,6 +55,24 @@ export interface DrawableGraph {
     readonly edges: Iterable<DrawableEdge>;
 
     /**
+     * selectedItems  
+     *   The set of selected graph elements.
+     */
+    readonly selectedItems: Set<Drawable>;
```
slaymaker1907: I did not know that you could make interfaces have readonly stuff. We should probably use this features of ts more often throughout the project.

app/components/graph-editor/drawable-interfaces.ts
```diff
@@ -61,6 +55,24 @@ export interface DrawableGraph {
...
+
+    /**
+     * origin  
+     *   The displacement of the origin point of the graph editor canvas.
+     */
+    origin: { x: number, y: number };
```
slaymaker1907: Maybe make this a Point interface that specifies an x and a y.

app/components/graph-editor/make-draw-node.ts
```diff
@@ -0,0 +1,111 @@
...
+
+
+/**
+ * makeDrawNode  
+ *   Makes a draw function for a given node.
+ */
+export function makeDrawNode(canvas: GraphEditorCanvas, g: CanvasRenderingContext2D, n: DrawableNode, dim: any, isDragging: boolean, isHovered: boolean, pt?: point): () => void {
+    // TODO:
+    // Deal with pt.
+    let shadowColor = (isDragging ? CONST.NODE_DRAG_SHADOW_COLOR : (isHovered ? CONST.SELECTION_COLOR : undefined));
```
slaymaker1907: It is kind of hard to read since there are nested terenary statements.

app/components/graph-editor/make-draw-node.ts
```diff
@@ -0,0 +1,111 @@
...
+                        break;
+                }
+                canvas.drawText(n.position, dim.th, lines, "#fff", 2, "#000");
+                canvas.drawCircle(pt, 5, "solid", 1, "#000", "#fff");
+            };
+        }
+        /////////////////////////////
+        // Labeled, Without Anchor //
+        /////////////////////////////
+        return () => {
```
slaymaker1907: The switch statement and drawText should be one thunk to avoid duplicate code. The only difference between the two returns in this function appears to be that the earlier one has a drawCircle in it.

app/components/graph-editor/README.md
ilargierdi: I like this idea. I also stole the idea of making a readme from Sheyne, so credit where credit is due. It made sense to make a readme, since the header comment in `graph-editor.component` was getting pretty long.

Sheyne:
Actually that was Daniels idea.

app/components/graph-editor/drawable-interfaces.ts
ilargierdi: I'm not sure if it's really enforced, but I put it there so tsc can hopefully catch me if I try to set it's value as opposed to just modifying the collection.

app/components/graph-editor/drawable-interfaces.ts
ilargierdi: I'm not sure I like the idea of exporting the `point` type from `canvas.ts`, but I see your... argument.

app/components/graph-editor/make-draw-node.ts
ilargierdi: I agree. I'm trying to keep the file short, though, so I don't have to scroll as much.

ilargierdi:
I sit corrected.

app/components/graph-editor/make-draw-node.ts
ilargierdi: I'll keep that in mind going forward.

Sheyne:
Interfaces are just aliases. So it makes no difference if we use an interface. As long as the fields match everything works out.

ilargierdi:
I assume you're responding to the `point` comment?

Sheyne:
Yeah

RighteousRaichu:
Just an FYI, zooming is waaaaay to touchy on my Mac. But we can work on making it smooth on all OS's later.

PR #73, opened 2017-01-30
formatting

Sheyne:


RighteousRaichu:
God damnit I was hoping you wouldn't notice this.

Sheyne:
If you noticed this, you could have put it in the PR.

The only reason I noticed was that I couldn't switch to master after I reviewed CJ's branch.

ilargierdi:
My bad.

PR #78, opened 2017-01-31
WindowService implements ModalService, new file dialog

RighteousRaichu:
@Sheyne you probably wanna focus your review at dynamic-component.component.ts where a component is dynamically loaded into the new window,
and modal-window.ts where what it means to be a ModalService is defined. (Currently WindowService is the only thing that implements this.) Theoretically we could make a none-electron version with a different ModalService and the main app would still work.


RighteousRaichu:
On macOS it creates a slide-down dialog... it was totes worth all the work for this.

Sheyne:
The slide-down dialog is SUPER cool. But it's a bit slow, any idea what causes the lag? Also the buttons look very out of place an Javaish.

app/components/main/main.component.ts
```diff
@@ -158,7 +168,7 @@ export class MainComponent implements OnInit, MenuEventListener, REPLDelegate, T
     menuEvent(e: MenuEvent) {
         switch (e) {
             case MenuEvent.NEW_FILE:
-                this.newFile();
+                this.promptNewFile();
```
Sheyne: Not sure why this isn't triggering for the first tab, but it should show this prompt at startup, or at least a blank screen with a message say please create a new document or load something.

app/components/main/main.component.ts
```diff
@@ -125,6 +127,14 @@ export class MainComponent implements OnInit, MenuEventListener, REPLDelegate, T
         this.selectedTab(tabNumber);
     }
 
+    promptNewFile() {
+        let [_, result] = this.windowService.createModal("sinap-new-file");
```
Sheyne: Create modal is AWESOME

RighteousRaichu:
The lag is because it needs to load angular in the new window and angular needs to bootstrap itself and compile the templates. If once we start using AOT it doesn't speed up significantly, we can preload the window as an optimization.

The entire dialog needs to go through and be made pretty so don't worry about the looks, it's sorta a glorified mockup right now.

RighteousRaichu:
(We would preload just one window, and the dynamic component loader could switch between components for different dialogs.)

app/index.ts
```diff
@@ -55,25 +55,98 @@ function createWindow() {
...
 //
 // Terminates the application when all windows have been closed.
 //
 app.on("window-all-closed", () => {
-    if (process.platform !== "darwin") {
-        app.quit();
-    }
+    //if (process.platform !== "darwin") {
+    //    app.quit();
+    //}
```
Sheyne: Just kill this, no need to leave the comment lying around. It's in the revision history.

app/index.ts
```diff
@@ -55,25 +55,98 @@ function createWindow() {
...
 // This will likely occur on mobile devices with limited resources.
 //
-app.on("activate", () => {
-    if (win === null) {
-        createWindow();
+// app.on("activate", () => {
+//     if (win === null) {
+//         createWindow();
+//     }
+// });
```
Sheyne: Same with this. No reason to leave around commented out code. Make note of the SHA of the commit that has it elsewhere and delete.

app/index.ts
```diff
@@ -55,25 +55,98 @@ function createWindow() {
...
+
+
+
+
+
+/** Managing Additional Windows **/
+// TODO: probs should split this into it's own file.
+
+var windows = new Map<Number, [Electron.BrowserWindow, ModalInfo]>();
```
Sheyne: I'm not sure I follow this bit, do we have support for several sinap root windows?

app/index.ts
```diff
@@ -55,25 +55,98 @@ function createWindow() {
...
+ipcMain.on('getWindowInfo', (event, arg: Number) => {
+    var window = windows.get(arg);
+    event.returnValue = window ? window[1] : null;
+});
+
+function createNewWindow(selector: string): Number {
+    if (win) {
+        var newWindow = new BrowserWindow({
+            parent: win,
+            modal: true,
```
Sheyne: Do we only support modal windows?

app/modal-windows/components/dynamic-component/dynamic-component.component.ts
```diff
@@ -0,0 +1,53 @@
...
+// Created by: Daniel James
+// Date created: January 30, 2017
+//
+
+
+import { Component, ViewContainerRef, ViewChild, ComponentFactoryResolver, ComponentRef, OnInit, Type } from '@angular/core';
+import { Title } from '@angular/platform-browser';
+import { WindowService } from './../../services/window.service';
+
+import { NewFileComponent } from './../../../components/new-file/new-file.component'; //TODO, shorter way to do this...?
```
Sheyne: You can remove the initial "./" not sure how much better than that you can do.

app/modal-windows/components/dynamic-component/dynamic-component.component.ts
```diff
@@ -0,0 +1,53 @@
...
+    template: `<div #container></div>`,
+    providers: [WindowService]
+})
+export class DynamicComponent {
+    private currentComponent: ComponentRef<any>;
+
+    @ViewChild('container', { read: ViewContainerRef })
+    private container: ViewContainerRef;
+
+    // TODO: I wish there was a better way to do this.
```
Sheyne: Decorators will let you register the components and their types to this dictionary automatically. 

(See how I do this in core.ts)

app/modal-windows/services/window.service.ts
```diff
@@ -0,0 +1,71 @@
...
+    public windowInfo: ModalInfo | null; // Main window will be null
+
+    constructor(private _ngZone: NgZone) {
+        ipcRenderer.on("windowResult", (event, arg) => this.callback(arg as ModalInfo));
+
+        // TODO: sending synchronous IPC here could be a bad idea, a better idea would be to send it async and store windowInfo as a Promise
+        this.windowInfo = ipcRenderer.sendSync("getWindowInfo", remote.getCurrentWindow().id);
+    }
+
+    public createModal(selector: string): [ModalInfo, Promise<any>] {
```
Sheyne: Really my only question is: Do we support any other kind of child windows? Do we support secondary main windows? (Now or ever)

app/components/main/main.component.ts
ilargierdi: Something like [this](https://github.com/2graphic/sinap-design-docs/blob/master/ui_sketches/blank_canvas.png)?

RighteousRaichu: Yeah that was a complete oversight on my part but I'll make it do that.

app/index.ts
RighteousRaichu: This supports multiple modal windows (though MacOS only displays one a time.) We can talk about support for multiple root windows but it's not a huge priority for me right now.

app/modal-windows/services/window.service.ts
RighteousRaichu: I think before merging I'll add in the ability to make it a modal window (slide down on Macs) or a child window (with it's own frame.) I think on other OS's they'll be the same.

app/models/modal-window.ts
```diff
@@ -10,10 +10,16 @@
...
 }
 
+export enum ModalType {
+    MODAL, // Will display as a drop down on macOS
+    CHILD, 
+} 
+
 export interface ModalService {
-    createModal(selector: string): [ModalInfo, Promise<any>];
+    createModal(selector: string, type: ModalType): [ModalInfo, Promise<any>];
```
Sheyne: I've found that a Promise<any> allowed Typescript to make some bad inference that ended up messing with my compiles. I tried using specific types (even if it was a type that in effect was just `BadType` so that it forced a cast and I had to think about what I was doing).

RighteousRaichu: Here when I use the any inside the Promise, because the any comes from an IPC call, there's no way to guarantee it'll be a certain type and I cast it to what I expect it to be.

Going forward I think we use Promises as expected and hope Typescript improves in this area

RighteousRaichu:
I've been looking into what it'll take to preload the windows, I think it's possible. If once I enable AoT and minification the startup time is still too slow, the approach I'll take is preloading modal windows and creating a class that manages all of that transparently.

Issue #79, opened 2017-01-31
`Core` design


Sheyne:
The issue addresses a couple of things.

1. All properties in the PropertyLists are mapped to human keys that contain capitalization and spaces, this is annoying for devs and will probably bug our plugin developers. 
2. Name shadowing: currently there are two name spaces, one for plugin keys one for sinap keys. I see two solutions to this:
     1. put them in the same name-space, if the overlap, have them share data and only show the plugin one in the properties panel. This is useful because plugin creators can opt-in to arbitrary information about nodes and edges, even if it's just cosmetic. This is bad because it is somewhat magical. 
     2. only share name space for keys defined in the built-in Node/Edge types. This means source, destination, label, and edges might be available, but no other properties. This is less magical, so that's good, but its also less flexible. 
3. computed properties
    1. certain properties should be able to be marked as computed (for now limited to string and numeric properties). If they're computed then they should be read only and be updated by some function every-time one of the properties they depend on changes. This is useful for making the label dependent on some other properties such as in the case of turing machines. 
    2. the scope questions above are particularly poignant with computed properties as it would allow the display properties of a node/edge to be computed based on its attributes. 

So if we enable **2.1**, how do we implement it? 

My idea is to add a property (typescript property not `PropertyList` property) to `Core.Element` called backer. The `@DrawableProperty` decorator would automatically convert all fields decorated to a getter/setter pair (this is apparently possible) and have them map to reads and writes on `Core.Element`'s `this.backer`. Plugin properties would also have a mapping from nice names to programmer names and would store their values ask keys on `backer` via their programmer names. Computed values would disable the setters for some of the Drawable/Plugin properties. Since everything is going through the getter/setter mechanism, we can detect events and determine the right time to recompute the computed properties. 

I need support for this soon. It is a blocker for the loading-types branch. 

Thoughts? Ideas?

Sheyne:
This pertains to everyone, so I'll tag you all: @slaymaker1907 @RighteousRaichu @ilargierdi 

I've implemented a lot of the above, but I have some thoughts on the way that matching of field names will occur. Classes will have two labels for each field: the camelCase name and the "Display Name". My question is how this should pertain to inheritance. The code below defines two classes (`A` and `B` where `B` inherits from `A`). What I'm thinking of doing is having (in the case of `A`) `Value` be the "true" name for the field and `val` be the shorthand used to reference it from code. So for all tasks involving inheritance `Value` is the bit that represents the symbol. 

[...]

Issue #63, opened 2017-01-25
Implement computed properties


Sheyne:
this will involve:

 - design
    - what syntax do we want to expose in the meta language
    -  @sheyne
 - implementation in type system
    - @sheyne
 - implementation in the core model
    - @RighteousRaichu

Issue #15, opened 2016-11-19
Plugin Node/Edge Property Specification


slaymaker1907:
Plugins should be able to define the set of properties for every node/edge. This fulfills part of issue #1

PR #88, opened 2017-02-02
Loading types from file

Sheyne:
Addresses issues: #79 #15 and lays groundwork for #63


Sheyne:
This definitely needs some work before it can be merged, but I'd like to get the discussion going. Also I'd like to begin implementing the Tools Panel. We can build that easily off this PR.

app/components/properties-panel/properties-panel.component.html
```diff
@@ -55,7 +55,13 @@
 
                             <template [ngIf]="pair[1].subtype(Type.Node)">
                                 <label  class="node-label">
-                                    {{selectedEntity[group[1]].get(pair[0]).drawableProperties.get('Label')}}
+                                    <!-- TODO: Figure out a better way to allow nodes to sometimes be null
+                                        
+                                        Specifically, this doesn't have a value yet, and this breaks a lot of 
+                                        promises that say that there is always a start state, that is, we can't
```
RighteousRaichu: Wait I don't think we should say there is always a start state. Or am I misunderstanding you? When you say optional do you mean an Optional<of> type or a more colloquial optional? Because I think start state should definitely be an Optional<Type.Node> property. (If it's too hard for that plumbing we can just assume all properties can be null and draw them accordingly in the property's panel)
It's fine to special case stuff here now, because I'm going to make this all generic soon.

app/services/plugin.service.ts
```diff
@@ -145,46 +107,46 @@ export class PluginService {
...
 
-    private makePlugin(kind: string): ConcretePlugin {
+    public makePlugin(kind: string): Promise<ConcretePlugin> {
         switch (kind) {
             case "dfa.sinap.graph-kind":
-                const plugin = new ConcretePlugin();
-                plugin.script = this.fileService.readFile("./build/plugins/dfa-interpreter.js")// TODO: Put real file in here.
-                    .then((code: string) => this.sandboxService.compileScript(code));
-                return plugin;
+                let defintions = this.fileService.readFile("./dfa-definition.sinapdef")
```
RighteousRaichu: Does this path work? I don't see anywhere where it'd be pulled into the webpack target.

RighteousRaichu: With Webpack you could bring it into the bundle with something like, `.readFile(require("file!./../dfa-definition.sinapdef")`

app/models/object.ts
```diff
@@ -0,0 +1,28 @@
...
+        if (sup instanceof Type.TypeVariable) {
+            throw "Invalid type" + type;
+        }
+        addProperties(props, sup, done);
+    }
+    return props;
+}
+
+
+export class Object extends MappedPropertyList {
```
RighteousRaichu: I know there's a case to me made for name purity, but I think it'd be better to name this SinapObject. (Just to avoid any possible confusion with Javascript objects.)

app/components/properties-panel/properties-panel.component.html
Sheyne: In the case of DFAs, we should guarantee that there is always a start state before we send it to the plugin if the plugin doesn't defines start state as optional. So sinap should be able to mark that there isn't one yet, so the graph isn't valid. The type constraints apply to graphs being sent to the plugin.

app/services/plugin.service.ts
Sheyne: It works.

app/models/object.ts
Sheyne: Yeah, I'm not sure about that. I do import it as SinapObject where I use it.

app/components/properties-panel/properties-panel.component.html
RighteousRaichu: Are we going to move start state to be a graph only property then? This leads to the question what's the best way to let the plugin define special drawing in the graph editor for a start state? Currently it looks like nodes and the graph have a start state property? If nodes have that property I don't think there's a good way in the type system to say one and only one needs to be a start state.

Sheyne: Moving towards only the graph has a start state, once that is possible in the GUI.

app/models/core.ts
```diff
@@ -161,12 +163,17 @@ function DrawableProperty(name: string, type: Type.Type) {
 export class MappedPropertyList implements PropertyList {
     properties: [string, Type.Type][] = [];
     propertyMap = new Map<string, string>();
-    constructor(properties: [string, string, Type.Type][], private backerObject: any) {
-        for (let ent of properties) {
-            this.properties.push([ent[0], ent[2]]);
-            this.propertyMap.set(ent[0], ent[1]);
+    constructor(properties: [string, string, Type.Type][], public backerObject: any) {
```
slaymaker1907: The type of properties is too complicated and should be a struct.

app/models/core.ts
```diff
@@ -178,7 +185,51 @@ export class MappedPropertyList implements PropertyList {
...
         this.backerObject[this.key(property)] = value;
     }
+
+    hide(mask: Set<string>) {
+        const masked = new Set<string>();
+        for (let i = this.properties.length - 1; i >= 0; i--) {
+            const prop = this.propertyMap.get(this.properties[i][0]);
+            if (prop && mask.has(prop)) {
+                masked.add(prop);
+                this.properties.splice(i, 1);
```
slaymaker1907: Not sure if it is worth the optimization, but you could check for contiguous sequences of things to delete to reduce the calls to splice.

slaymaker1907: Another idea would be to just create a properties array to avoid the potential O(n^2) behavior.

app/models/core.ts
```diff
@@ -178,7 +185,51 @@ export class MappedPropertyList implements PropertyList {
...
+            }
+        }
+        return masked;
+    }
+}
+
+class MaskingPropertyList implements PropertyList {
+    maskedElements: Set<string>;
```
slaymaker1907: Did you intend for this to be public?

app/models/object.ts
slaymaker1907: I agree with @RighteousRaichu since Object is a builtin. If it would simply conflict from another module in the project whose name is not imported, that would be fine (i.e. Graph and other generic names).

app/services/plugin.service.ts
```diff
@@ -2,139 +2,101 @@ import { Injectable, Inject } from '@angular/core';
...
+    object: SinapObject;
+    constructor(public kind: string, t: Type.ClassType) {
+        this.object = new SinapObject(t, this.backer);
     }
 }
 
-class ConcretePlugin implements Core.Plugin {
-    kind = "dfa.sinap.graph-kind";
-    script: Promise<Script>;
+type Definitions = { all: Map<string, Type.Type>, nodes: Map<string, Type.ClassType>, edges: Map<string, Type.ClassType>, graphs: Map<string, Type.ClassType> };
```
slaymaker1907: Please put this on multiple lines, I had to scroll on GitHub while reviewing this to see the whole thing.

app/services/plugin.service.ts
```diff
@@ -2,139 +2,101 @@ import { Injectable, Inject } from '@angular/core';
...
+        if (!type) {
+            throw "type not found";
         }
+        return new PluginPropertyData(kind, type);
     }
-
-    edgePluginData(type: string) {
-        return new PluginPropertyData("Edge", []);
+    nodePluginData(kind: string) {
+        const type = this.definitions.nodes.get(kind);
```
slaymaker1907: nodePluginData et. al. should have a hidden private method that accepts kind and a Map since they are basically the same and seem to differ only in which dictionary is used.

app/services/plugin.service.ts
```diff
@@ -193,6 +155,42 @@ export class PluginService {
...
+
+        const nodes: [string, Type.ClassType][] = [];
+        const edges: [string, Type.ClassType][] = [];
+        const graphs: [string, Type.ClassType][] = [];
+
+        for (const [name, value] of scope.definitions.entries()) {
+            if (value.subtype(Type.Node)) {
+                nodes.push([name, value as Type.ClassType]);
+            }
+            if (value.subtype(Type.Edge)) {
```
slaymaker1907: Can things be both a Node and an Edge?

app/services/sandbox.service.ts
```diff
@@ -17,14 +17,12 @@ export class Script {
...
-        return new Promise((resolve, reject) => {
-            try {
-                resolve(this.nodeScript.runInContext(context));
-            } catch (err) {
-                reject(err);
-            }
-        });
+    runInContext(context: Context | Promise<Context>): Promise<NativeResult> {
+        if (!(context instanceof Promise)) {
+            context = Promise.resolve(context);
```
slaymaker1907: I feel like it would be cleaner to create a lambda above that handles a context. If it is a promise, pass that lambda into then. If it is not a promise, just run it directly since runInContext returns a promise.

Sheyne: I think this is cleaner, but I'm open to changing it.

app/services/plugin.service.ts
Sheyne: They *could* be. Not sure how that's useful, but I cant think of a reason to preclude it.

app/services/plugin.service.ts
Sheyne: I agree. This whole area needs a cleanup. Its loose strings hacked together.

app/services/plugin.service.ts
Sheyne: I agree. This whole area needs a cleanup. Its loose strings hacked together.

app/models/core.ts
Sheyne: No.

app/components/main/main.component.ts
```diff
@@ -217,9 +224,11 @@ export class MainComponent implements OnInit, MenuEventListener, REPLDelegate, T
                                 throw "invalid file format version";
                             }
 
-                            this.newFile(filename.substring(Math.max(filename.lastIndexOf("/"),
-                                filename.lastIndexOf("\\")) + 1),
-                                new Core.Graph(this.pluginService.getPlugin("dfa.sinap.graph-kind")));
+                            this.pluginService.getPlugin("dfa.sinap.graph-kind").then((plugin) => {
```
Sheyne: This should **not** be in a constants file. These are strings that will eventually be loaded based on the plugin.

Sheyne:
The constant duplication isn't actually an issue, because the constants will live outside of Sinap. Plugin  identifiers will be loaded from a dictionary somewhere.

app/components/main/main.component.ts
```diff
@@ -83,6 +86,8 @@ export class MainComponent implements OnInit, MenuEventListener, REPLDelegate, T
 
     onContextChanged = () => { // arrow syntax to bind correct "this"
         if (this.context) {
+            this.context.graph.activeEdgeType = "DFAEdge";
```
Sheyne: This is a simulation of the tools panel, so it has to be hard coded until that is built. It could be more realistic by getting the list of possible edge types and setting it to the first one.

Sheyne: By the way @RighteousRaichu, this is the thing to hook up to the tools panel.

app/models/core.ts
```diff
@@ -79,7 +80,8 @@ export class Graph extends Element {
     exampleTuple = ["hello", 10];
 
     constructor(public plugin: Plugin) {
-        super(plugin.graphPluginData());
+        // TODO: fix that this is obviously special cased
+        super(plugin.graphPluginData("DFAGraph"));
```
Sheyne: Remember what it was before? This is the only way to avoid duplication of a bunch of strings. What is your proposal?

slaymaker1907:
I think even if it is not necessary long term, we should put it in a variable in the short term since it seems like a very dangerous case of string duplication. Plus, it will make it easier to figure out where we are special casing things when the final refactor is done.

app/components/main/main.component.ts
slaymaker1907: See my reply below on this.

app/models/core.ts
slaymaker1907: It seems like Element could be an interface instead of a class. I'm not seeing an explosion of duplicate code by making Element an interface.

Personally, I think that the inheritance is a bigger issue than the annotations. I just think that the two combined spells for complicated code. Imagine Java Spring combined with Swing.

app/services/plugin.service.ts
slaymaker1907: It might be a good idea then to put in a comment stating as such so that no one changes it to an else if block.

app/services/plugin.service.ts
```diff
@@ -2,139 +2,104 @@ import { Injectable, Inject } from '@angular/core';
...
 }
 
 @Injectable()
 export class PluginService {
+    private plugins = new Map<string, Promise<ConcretePlugin>>();
     private interpretCode: Script;
     private runInputCode: Script;
-    private plugins: Map<string, ConcretePlugin> = new Map<string, ConcretePlugin>();
+    // TODO: load from somewhere
+    private pluginKinds = new Map([[MagicConstants.DFA_PLUGIN_KIND, { definitions: "./dfa-definition.sinapdef", interpreter: "./build/plugins/dfa-interpreter.js" }]])
```
RighteousRaichu: So I've been looking into why this is actually working. When we run `npm run start` the root path for electron is the current directory. This actually currently breaks if we do`npm run package` and run the program that way because the root path is different and this file isn't included in the target (ie this file isn't copied to the build directory.)
This is currently fine because these are just hardcoded and it's easiest this way, but when @slaymaker1907 starts messing around with loading dynamically it might be nice to know that `app.getAppPath()` will give you the path to where everything is.

If we ever do want to include files like this for whatever reason in the future it'd need to look something like,
`app.getAppPath()+require("file!dfa-definition.sinapdef");`

(Where `app` is app brought in by Electron's `remote` module.)

slaymaker1907: I don't think it will be necessary @RighteousRaichu since it will save the location of files to some sort of user data location.

RighteousRaichu:
@Sheyne do you think this is ready to be merged yet?

Sheyne:
Yeah

Issue #75, opened 2017-01-30
Clicking on "anchor" creates a self loop


Sheyne:
I'm not sure if this is desirable in general, but with un-selectable self loops, this is really annoying. I keep trying to select a node and adding a self loop.

Sheyne:
Maybe only show anchor when outside of node radius and keep the click for self loop behavior (which I think is good).

ilargierdi:
Just to clarify: you want the edge threshold of a node to be outside of the
actual node geometry such that the anchor (as well as edge creation) will
only be active if the cursor is outside of the node dimensions?

Sheyne:
Yep.

ilargierdi:
This behavior is now implemented in ff758df.

ilargierdi:
Merged as part of #89.

Issue #74, opened 2017-01-30
Zoom centers around mouse


Sheyne:
Currently the graph editor zooms from the top left, it should zoom from the location of the mouse

ilargierdi:
Currently implemented in eeaa623.

ilargierdi:
Merged as part of #89.

Issue #71, opened 2017-01-30
Flipping coloring on edge moves


Sheyne:
Currently a shadow is shown where the new edge will be and the old edge is left intact. This doesn't match the behavior of edge creation where the new edge has a different styling. 

I'd like consistant styling between created edges and moved edges and that the old edge become shadowed or in some way greyed out so it's clear which edge is getting moved. 

Proposal:

1. Created edges have the default styling @ 70% opacity
2. Moved edges keep their existing styling @ 70% opacity
3. Why an edge is being moved, the old location is kept, but at 50% opacity. 

What does everyone else think?

Whether or not we go with my suggestions, lets close this before Beta.

ilargierdi:
I think what you describe is how the drawing behavior was originally implemented. It was probably lost during the refactor. I would like to put this in effect, but I'll wait for everyone else to respond before putting it in.

RighteousRaichu:
Why 70% opacity instead of just 100%?

ilargierdi:
@RighteousRaichu Which edge are you talking about? Currently, if you want to move an edge from one node to another, 2 edges are drawn: the original one (which currently is drawn at 100% opacity) and the one being dragged (which has the same properties as an edge being created).

RighteousRaichu:
Sorry I was talking about the proposal. I think new edges or edges being moved should be drawn exactly as they will be (100% opacity) If an edge is being moved then where the old edge was(is?) should be slightly ghosted.

Sheyne:
Maybe visual feedback they're not yet valid.

ilargierdi:
This is currently implemented on 7ade31d.

ilargierdi:
Merged as part of #89.

Issue #57, opened 2017-01-20
Update hit detection.


ilargierdi:
Hit detection needs to be updated for the new bezier edges. Either use the technique of mapping drawables to hit functions or utilize the native canvas hit regions.

ilargierdi:
This now works in 7ade31d.

ilargierdi:
Merged as part of #89.

PR #89, opened 2017-02-02
Issue57

ilargierdi:
The code is currently a mess, but I wanted to push the new features onto master. There is one bug  that I'm working on, but it's taking more time than I'd like to spend on it. It involves the selection box and cubic bezier curves.

I fully intend to do some code maintenance and documentation on the graph editor stuff, but for now I'd like to take a break from working on it.

This PR addresses #42, #57, #71, #74, #75, and partially #41.

### DIFF
--- a/app/components/graph-editor/README.md
+++ b/app/components/graph-editor/README.md
@@ -1,5 +1,14 @@
+_Last updated: January 31, 2017_
+
 # Notes
-## Drawing Behavior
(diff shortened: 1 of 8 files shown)


ilargierdi:
I can change it back or increase the threshold. I think the idea is if you see an anchor point, it's safe to drop the edge.

PR #93, opened 2017-02-02
Edge creation region tuning.

ilargierdi:
This is mostly a feature request (and a minor bugfix), but I'd like some feedback on the behavior of the edge region around the node before merging into master.


Sheyne:
Actually, can you make it to that moving existing edges has higher priority than creating a new one? Its hard to grab the end of an edge now without triggering the creation behavior. I don't consider this a blocker though.

ilargierdi:
For dragging edges, you don't actually have to be anywhere near the anchor point to grab it. You could be close to the midpoint of the edge, and depending on which side of the midpoint you are on, you'll automatically start dragging whichever side the anchor point is being displayed. Making the changes you're requesting will take a bit of finesse, since if existing edges are going to take priority over hit detection, the edge threshold will have to be lowered enough so that you won't accidentally select the edge when you want to select the node. (The edge threshold currently overlaps past the center of the minimum node size).

ilargierdi:
I could make it happen, though. But it'll have to be after all of the other glaring issues with the codebase are handled. :p

Sheyne:
So should I make an issue for it? And for now I'll just tone down the outer radius a little.

ilargierdi:
Yeah, sounds good.

ilargierdi:
I think what you're asking is to shrink the inner threshold of the node for edge creation. If the cursor is far enough inside the node, the anchor should not appear unless you're dragging an edge over the node. The last PR for the graph editor made it so that there was no inner threshold on the node. I think what Sheyne originally wanted was to just shrink the inner threshold as opposed to removing it entirely.

RighteousRaichu:
No I'm talking about moving edges. If you hover over an existing edge, depending on which side the mouse is closest to an anchor is displayed. I think this might make people think that they have to click on the anchor to drag the edge, when clicking the anchor will actually make a new edge.

> Its hard to grab the end of an edge now without triggering the creation behavior.

That's the comment from Sheyne I was talking about.

ilargierdi:
I see what you mean. The anchors in that case are to indicate which side of the edge will be dragged if the user starts dragging the edge. Removing them might take a bit of work, but then the user wouldn't have any visual indication as to which side of the edge is going to dragged. Changing the visual indication would likely take an equal amount of work.

RighteousRaichu:
I vote we either remove them or change them to a different visual indicator. I just can't think of a good way to do it though. For now it's probably fine.

Sheyne:
I think they're actually pretty good. The problem is that as you approach
the node, even if you're fairly close to the edge if you switch to "create
mode" from move mode, even if you're hovering on the edge itself. So it
might just be swapping the order of if statements.

RighteousRaichu:
@slaymaker1907 I'm going to merge this soon unless you have any objections.

RighteousRaichu:
@ilargierdi do you want me to merge this or did you want to discuss it more?

ilargierdi:
You're free to merge, unless there's anything that immediately needs to be changed about it.

PR #94, opened 2017-02-03
Move the typesystem as an external dependency

Sheyne:
I'll be moving out other core parts soon


ilargierdi:
Fancy. I kind of want to do this with the graph editor; it'll definitely make testing easier.

Sheyne:
I did it because the CLI will also depend on it.

Sheyne:
This ended up merged in with 2graphic/sinap-ide#loading-types

PR #97, opened 2017-02-03
Make sure the hidden div actully gets hidden.

RighteousRaichu:
So I experimented with using *ngIf instead of [hidden], however that didn't play nice with the @viewChild annotations and would've been an ugly mess.


ilargierdi:
I think the way you ended up doing it is fine.

PR #106, opened 2017-02-07
Removed promises and fixed compilation messages.

slaymaker1907:
Now, plugin developers don't have to use Promises, compilationMessages works again, and graphs with loops are correctly passed to the input programs.

This is a somewhat in flux change with the changes that are to be implemented with the plugin type system, but I think that that this code is already quite useful and will form a better base with serializing graphs. In particular, now plugins can throw exceptions and have those displayed as output which will be a huge productivity increase.

I've added @ilargierdi to this PR so that he can see the developments with plugins if he so chooses as well as to get another pair of eyes on it.


app/models/plugin.ts
```diff
@@ -20,7 +12,7 @@ export type ProgramOutput = string | boolean;
 /**
  * Represents a function which converts graphs into programs or else returns an error that has toString.
  */
-export type Interpreter = (graph: Graph) => Promise<Program>;
+export type Interpreter = (graph: any) => Promise<Program>;
```
ilargierdi: Will this ever be changed to an actual type?

plugins/dfa-interpreter.ts
```diff
@@ -1,95 +1,207 @@
...
 }
 
-class Graph {
-    edges: Edge[];
-    nodes: Node[];
+/**
+ * This represents a compiled program given an input computation graph. It should be immutable, though the RunningProgram returned by initDebugging may be mutable itself.
+ * If desired, a simple run method or initDebugging method can be provided and then fillInProgram will fill out the rest of the required fields/methods.
+ */
+export interface Program {
```
ilargierdi: I the future, I think these interfaces and type aliases should be in their own dedicated file. It looks to me like these are meant to be part of the plugin API that defines the contract for what a valid Sinap interpreter should be.

app/models/plugin.ts
slaymaker1907: It will be changed from any once serialization is reworked after the plugin typing update. This type is supposed to be a serialized graph.

plugins/dfa-interpreter.ts
slaymaker1907: I definitely agree. I mostly just shoved it into one file out of laziness as well as the fact that it was subject to change extensively with the updates that Sheyne is working on.

PR #108, opened 2017-02-07
Unit test panel - WIP

RighteousRaichu:
There's still a huge problem where the last test that's run in quick succession is the result for all the tests.
Another problem which I need to solve before this will really be finished in some sane change auditing (also necessary for undo/redo) so the unit tests panel can know if it's up to date or inform the user if it isn't, but I'm trying my best here to stay up to date.
Also this is not the final look and feel of things.
(Input/Output types need to be generic types, also, obviously.)


app/components/test-panel/test-panel.component.ts
```diff
@@ -4,16 +4,105 @@
...
+        });
+    }
+
+    private runTest(test: Test) {
+        if (this.program) {
+            this.program.run(test.input).then((output) => {
+                console.log(test, output);
+                test.output = output as boolean;
+            }).catch((error) => {
+                test.output = null;
```
slaymaker1907: You should set test.output = error instead of swallowing the error.

slaymaker1907:
It seems like there is some sort of race condition going on with running a program. Ideally these will be fully asynchronous anyway, but reusing context is inherently not thread safe so I'll probably need to use some sort of locking mechanism.

Issue #84, opened 2017-02-01
Menu Bar shows up in the New File Dialog on Windows.


ilargierdi:
See slack for an image.

Issue #51, opened 2017-01-11
Fix copy/paste.


RighteousRaichu:
Copy/paste shortcuts behave weird within electron. Without menu items for these, the functionality doesn't work.

Sheyne:
That actually true for all Mac apps.

slaymaker1907:
Seems to work for Windows, but this bug happens in MacOS. Not tested in Linux.

ilargierdi:
For reference: https://developer.mozilla.org/en-US/docs/Web/Events/copy

PR #109, opened 2017-02-08
Changes related to the menu.

RighteousRaichu:
* #54 Fixed shortcut for Developer Tools on Mac... also only opens for development build.
* Make Webpack not replace Electron's process with Node's process.
* #3 Removed 'Sinap' menu item from non macOS targets (shows as Electron if ran from the command line.)
* Moved the menu template to models/menu.ts and made it a lot more robust with the type system and etc...
* #51 Fixed Copy/Paste/Undo/Redo on macOS
* Stubbed out public copy, cut, paste methods in the graph editor. These get called only if the editor has focus, otherwise they have their default, OS defined, actions.
* #84 Now modal windows don't have a menu (Someone check this please.) 

I tagged you all as reviewers to just make sure this hasn't broken something in Linux/Windows, pls check, thx.


app/components/graph-editor/graph-editor.component.ts
```diff
@@ -334,6 +334,18 @@ export class GraphEditorComponent implements AfterViewInit {
...
+        console.log("Editor copy");
+    }
+
+    cut(): void {
+        console.log("Editor cut");
+    }
+
+    paste(): void {
+        console.log("Editor paste");
+    }
```
ilargierdi: These might not be necessary in the editor once the new implementation is rolled out. You have access to the collection of selected entities from the `DrawableGraph` object now as well as its validator, create, and delete methods.

RighteousRaichu: I still think it probably makes sense for the graph editor to be in charge of cut, copy, paste itself.  The main component is more of a director class and the nitty gritty implementation of stuff should be outsourced, in this case either to graph editor itself or something else.

RighteousRaichu:
I added a commit to fix a bug and enable zooming in the menu (works great.)
Thanks for checking on Window's @ilargierdi

PR #110, opened 2017-02-10
Webpack 2, production builds, webpack DLL support, remove windows specific npm scripts.

RighteousRaichu:
* Upgrade from webpack 1.x -> 2.x
* Specify a special "Harmony" fork of 'uglifyjs', which supports ES6. Will require you to clean out 'node_modules' and 'npm install' again. This makes our generated code about 3-4x smaller.
* Got rid of windows specific npm scripts, using `rimraf` and `cpx` instead of `rm` and `cp`
* Cleaned up webpack config file, and provide an option for a production build.`npm run package` runs a production build.
* Using faster to generate source-maps. They still work well enough for me, but you can change to better sourcemaps my modifying the webpack configuration to `devtool: ‘source-map’`
    * Just incase you don’t already know, you can open the dev tools then open the `sources` tab, then in the left hand column `webpack -> . -> app -> our sources` to select and view the sourcemapped files and set break points.
* Using the webpack DLL plugin to build our vendor bundle. This means you only have to build this once after updating angular. `npm run build:dll` the regular build expects this to exist, but now doesn’t need to build angular everytime. For now the dll is always uglified with no source maps. Build no longer runs `npm install` automatically

Example build times on my computer:
* `npm run build:dll` -> ~16 seconds
* `npm run build` -> ~23 seconds (3-4ish seconds for tsfmt)
* `npm run build:prod` -> ~29 seconds
* Incremental changes with `npm run build:watch` -> <1 second
* build directory size for production build -> ~1.4MB.
* build directory size for development build -> ~4.1MB. (Was a bit higher before all these changes.)
* `npm run package` -> ~119 seconds
* `npm run build` before DLL -> ~38 seconds

As a summery to build you’ll want to:
* Run `npm run clean:node_modules` once.
* Run `npm install` once now and once anytime you want to update dependencies.
* Run `npm run build:dll` once now and once anytime you update angular (or once anytime after running `npm run clean`)
* Finally for most changes just `npm run build` or `npm run build:watch` and `npm run start` like usually. (or `npm run build:watch:start`)

Can we make sure these changes work for everyone before merging to master? Thanks!


ilargierdi:
I'm getting build errors on my Windows machine, even from a fresh clone. The error message says it's coming from line 15:
```javascript
module.exports = (env = {}) => { // pass command line arguments like `webpack ... --env.arg=value`
                      ^

SyntaxError: Unexpected token =
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at requireConfig (C:\Users\CJ\Documents\Projects\git\sinap-ide\node_modules\webpack\bin\convert-argv.js:96:18)
    at C:\Users\CJ\Documents\Projects\git\sinap-ide\node_modules\webpack\bin\convert-argv.js:109:17
    at Array.forEach (native)
```
Also, can we change `clean:node_modules` to just `clean:node`? And I think it would be worthwhile to update the readme with the build summary from OP.

RighteousRaichu:
I'll update the readme that's a good idea. Also I was just going to remove the `clean:node_modules` script since rimraf doesn't like deleting itself.

For the error, what version of node are you using? And maybe try updating that and see if that fixes the error. (It worked fine on Dyllon's Surface so it's not a window's specific issue.)

ilargierdi:
Looks like I had an older version of node. Upgrading fixed the build issues.

PR #111, opened 2017-02-11
Implemented Better File System API

slaymaker1907:
This does not actually completely fix saving so this branch should not be deleted just yet, but I thought the changes were significant enough to pull into master at this stage.

Only review if you wish to. I thought that these changes were pretty cross cutting across the project and so I included everyone.


webpack.config.js
```diff
@@ -25,7 +25,6 @@ module.exports = (env = {}) => { // pass command line arguments like `webpack ..
             __dirname: false,
             __filename: false,
             fs: false,
-            process: false,
```
RighteousRaichu: Wait we need this line here, or to figure out a way to make it work for both of us.

RighteousRaichu: Actually nevermind it seems to be fine, so we're good.

Issue #107, opened 2017-02-07
Properties don't have default values


Sheyne:
`Accept State` and `Start State` are booleans, but you have to toggle them on off or they get sent to the interpreter as undefineds. I'm working on figuring this out in the general case.

RighteousRaichu:
Related to #91

RighteousRaichu:
@Sheyne I imagine we could do this in graph controller, however, should this be done in core?

For Beta we have just made the interpreter a little more verbose so you don't have to toggle the states in the IDE, so I'm moving this to 1.0.

RighteousRaichu:
Being done in sinap-core and/or sinap-types

Issue #103, opened 2017-02-06
Editor graph representation


ilargierdi:
### TL; DR: Should `GraphDelegate` have factory methods for creating nodes and edges?

As discussed earlier today, we agreed to give drawables a default implementation as opposed to being interfaces. The `DrawableGraph` was proposed to have its constructor take a `GraphDelegate` that would supply at the very least a validating method for edges. I do like the name `isValidEdge` better than `canCreateEdge`, but that is just a minor detail.

What my main concern for opening this discussion is what other methods to put on the `GraphDelegate` interface. You mentioned that it needs to have some notion of `activeNode` and `activeEdge` for creating nodes and edges. I'm not sure that the `DrawableGraph` should implement the creation logic for anything other than having wrapper methods to call something on the `GraphDelegate` for both validation and creation. The `GraphDelegate` would then be a factory and validator for graph elements in this case.

The issue I have with pushing graph element creation onto the `DrawableGraph` is that the class will need prior knowledge about `kind`s of nodes and edges and how to set their properties. This boils down to type casing on nodes and edges, which would have to be supplied by a third party, anyway.

The only concern the graph editor should have is drawing behavior, which are supplied by drawable properties. This was the main idea behind `DrawableEdge` and `DrawableNode`; they are literally just objects with properties and no functionality from the point of view of the editor. The editor does, however, need to be able to query something for new graph elements; it just doesn't need to know how to create them. I liked your proposal for a `GraphDelegate` interface, and I also like the idea of a factory object to create nodes and edges. So my question is: would it be reasonable to include factory methods on `GraphDelegate`?

Sheyne:
You're right. The graph editor doesn't need any notion of an active node/edge type. Since that's the case, maybe we should just take a single `validator: (Node, Node) => boolean` argument.

ilargierdi:
The editor still needs to be able to request new elements, though. I think supplying a `GraphDelegate` or `GraphFactory` object would be better. Maybe the factory methods could take a drawable as its parameter or maybe they take no arguments and return a drawable. Otherwise, the editor component could emit events to indicate when a graph element should be created or destroyed and the third party hooks into that, but that doesn't feel right.

ilargierdi:
Or we could do something like Angular2's dependency injection, but that seems like it might take more effort.

Sheyne:
So we've got a couple of things to tackle:

1. graph editor creates nodes and edge
2. graph editor wants to validate that an edge is valid
[...]

Issue #41, opened 2016-12-08
Selection box hit testing needs to work on any part of any component.


ilargierdi:
Currently the selection box only works on nodes and only selects a node if the center of the node is in the selection box. Any edges connected to the selected node are also added to the selection set.

The selection box needs to be able to add components to the selection set if any part of a component is within the select region. E.g. if an edge passes through the select region, it should be added to the selection set.

ilargierdi:
This is now partially implemented. Need to nail down intersections with cubic beziers.

ilargierdi:
#120

PR #120, opened 2017-02-17
Model Updates/Graph editor maintenance

Sheyne:
- Restructures the internal model. 
    - Graph editor now keeps its own model of the graph
    - Creates main-graph.ts which syncs `CoreElements` with the graph-editor's model
 - Adds support for the new version of sinap-core.
 - Fixes reusing plugin contexts
 - Reimplements saving/loading in a future proof manner


ilargierdi:
Related: #103 and #41

app/components/graph-editor/defaults.ts
```diff
@@ -99,20 +97,17 @@ export const NUDGE: number = 3;
...
-export const NODE_DEFAULTS: DrawableNode = {
-    position: { x: 0, y: 0 },
-    label: "",
-    shape: "circle",
-    color: "#fff200",
-    borderColor: "#000",
-    borderStyle: "solid",
-    borderWidth: 1
+ */
+export const NODE_PROPERTIES = {
```
slaymaker1907: I assume that this are get to avoid mutation, but that seems kind of silly to me. I think it would be more readable as a JSON object constant.

app/components/graph-editor/drawable-element.ts
```diff
@@ -0,0 +1,275 @@
+// File: drawable-element.ts
+// Created by: CJ Dimaano
+// Date created: February 4, 2017
+
+
+import { FONT_SIZE } from "./defaults";
+import { GraphEditorCanvas, point, rect, size } from "./graph-editor-canvas";
+import { Drawable } from "./drawable";
+import { DrawableGraph } from "./drawable-graph";
+import {
```
slaymaker1907: We should decide on multline imports or single line imports to have consistent style.

app/components/graph-editor/drawable-element.ts
```diff
@@ -0,0 +1,275 @@
...
+    Default,
+    Dragging,
+    Hovered
+};
+
+/**
+ * DrawableElement  
+ *   Common properties shared between drawable edges and nodes.
+ */
+export abstract class DrawableElement extends Drawable {
```
slaymaker1907: This seems pretty complicated. Would it be possible to avoid an abstract class by pulling this out into an interface and a putting the logic into a common class/function?

app/components/graph-editor/events.ts
```diff
@@ -0,0 +1,60 @@
+// File: events.ts
+// Created by: CJ Dimaano
+// Date created: February 7, 2017
+
+
+export type Listener<A extends EventArgs> = (evt?: A) => void;
+
+export type PropertyChangedEventListener<T> = Listener<PropertyChangedEventArgs<T>>;
+
+export class EventArgs {
```
slaymaker1907: Is there a reason why this is a class instead of an interface?

app/components/graph-editor/events.ts
```diff
@@ -0,0 +1,60 @@
...
+    constructor(public readonly source: any) { }
+}
+
+export class CancellableEventArgs extends EventArgs {
```
slaymaker1907: Same thing about why is this a class instead of an interface?

app/components/graph-editor/events.ts
```diff
@@ -0,0 +1,60 @@
...
+        public readonly curr: T
+    ) {
+        super(source);
+    }
+}
+
+export class EventEmitter<A extends EventArgs, L extends Listener<A>> {
+    protected listeners: Set<L>
+    = new Set<L>();
+    addListener(l: L) {
```
slaymaker1907: This parameter name looks like the number one and should be made more readable.

app/components/graph-editor/events.ts
```diff
@@ -0,0 +1,60 @@
...
+    emit(args: A) {
+        this.listeners.forEach(v => v(args));
+    }
+}
+
+export class CancellableEventEmitter<A extends CancellableEventArgs, L extends Listener<A>>
+    extends EventEmitter<A, L> {
+    emit(args: A) {
+        for (const l of this.listeners) {
+            l(args);
```
slaymaker1907: This parameter should also be renamed for readability.

app/models/double-map.ts
```diff
@@ -0,0 +1,49 @@
+
+
+/**
+ * Keep two dictionaries in sync, different efficient maps to the same data
+ */
+export class DoubleMap<A, B, C> {
```
slaymaker1907: If this is only used one place, then make it non-generic since generic code is inherently more complicated.

app/models/main-graph.ts
```diff
@@ -0,0 +1,336 @@
...
+ * 
+ * Calls to get/set will update both. It also conatins a 
+ * `proxy` field. Reads and writes to this will intelligently update 
+ * the core element and the drawable. For any of these uses, if 
+ * the backer object (core/drawable) references another core or drawable
+ * this will return a `BridgingProxy` in its place. Sets will intellegently
+ * sync the backends even if the set `v` is a `BridgingProxy`
+ */
+export class BridgingProxy {
+    proxy: { [a: string]: any };
```
slaymaker1907: We should have a mappable interface or something for this for readability.

app/models/main-graph.ts
```diff
@@ -0,0 +1,336 @@
...
+
+    set(k: PropertyKey, v: any, updateDrawable = true) {
+        let drawableValue = v;
+        let coreValue = v;
+
+        // If v is a drawable element or a core element,
+        // then when it gets set in the various proxied data
+        // structures, set it differently. 
+
+        if (v instanceof BridgingProxy) {
```
slaymaker1907: Could we change BridgingProxy, potentially adding in a second class, so that we don't have to do instanceof checks?

app/models/main-graph.ts
```diff
@@ -0,0 +1,336 @@
...
+        const coreValue = this.core.data[k];
+        // if this is something that should be proxied
+        const proxiedValue = this.graph.bridges.getB(coreValue);
+        // return the proxy. That is, any changes made to the returned 
+        // object should propogate to both backers
+        return proxiedValue !== undefined ? proxiedValue : coreValue;
+    }
+}
+
+function drawableFromAny(a: any, bridges: DoubleMap<Drawable, CoreElement, BridgingProxy>) {
```
slaymaker1907: This really seems like it should be a method on bridges.

app/models/main-graph.ts
```diff
@@ -0,0 +1,336 @@
...
+export class UndoableAdd {
+    constructor(public bridge: BridgingProxy) { }
+}
+
+export class UndoableDelete {
+    constructor(public bridge: BridgingProxy) { }
+}
+
+// TODO: changes can also be adds or deletes
+export class UndoableChange {
```
slaymaker1907: What if we just made this an interface with two methods, undo and redo?

app/models/main-graph.ts
```diff
@@ -0,0 +1,336 @@
...
+    constructor(public target: BridgingProxy,
+        public key: PropertyKey,
+        public oldValue: any,
+        public newValue: any) { }
+}
+
+export class MainGraph {
```
slaymaker1907: This should be renamed to GraphController.

app/models/main-graph.ts
```diff
@@ -0,0 +1,336 @@
...
+    drawable: DrawableGraph;
+    activeNodeType: string;
+    activeEdgeType: string;
+    public changed = new EventEmitter<UndoableEvent>();
+
+    private selectedElements: Set<BridgingProxy>;
+
+    public bridges = new DoubleMap<Drawable, CoreElement, BridgingProxy>();
+
+    constructor(public core: CoreModel, public plugin: Plugin) {
```
slaymaker1907: It seems like these should be readonly or have an explicit setter.

app/services/plugin.service.ts
```diff
@@ -1,236 +1,61 @@
...
-    }
-    edgePluginData(kind: string) {
-        const type = this.definitions.edges.get(kind);
-        if (!type) {
-            throw "type not found";
-        }
-        return new PluginPropertyData(kind, type);
-    };
+export declare class Program {
+    constructor(any: SerialJSO);
```
slaymaker1907: We should rename this to avoid confusing it with the type any.

app/services/plugin.service.ts
```diff
@@ -1,236 +1,61 @@
...
+    private plugins = new Map<string, Plugin>();
+    private programs = new Map<Plugin, Promise<StubContext>>();
+    private getResults: Script;
+    private addGraph: Script;
     // TODO: load from somewhere
-    private pluginKinds = new Map([[MagicConstants.DFA_PLUGIN_KIND, { definitions: "./dfa-definition.sinapdef", interpreter: "./build/plugins/dfa-interpreter.js" }]])
+    private pluginKinds = new Map([[MagicConstants.DFA_PLUGIN_KIND, "./plugins/dfa-interpreter.ts"]])
 
-    constructor( @Inject(LocalFileService) private fileService: FileService,
+    constructor( @Inject(LocalFileService) private fileService: LocalFileService,
```
slaymaker1907: Why was this changed to use LocalFileService instead of FileService?

app/services/plugin.service.ts
```diff
@@ -1,236 +1,61 @@
...
-        if (loadedPlugin) {
-            return loadedPlugin;
+    public getPlugin(kind: string) {
+        let plugin = this.plugins.get(kind);
+        if (plugin) {
+            return plugin;
+        }
+        const fileName = this.pluginKinds.get(kind);
+        if (!fileName) {
+            throw "Plugin not installed";
```
slaymaker1907: Make sure to specify which plugin kind in the error message.

app/services/plugin.service.ts
```diff
@@ -1,236 +1,61 @@
...
-            }
-            if (value.subtype(Type.Edge)) {
-                edges.push([name, value as ClassMetaType]);
-            }
-            if (value.subtype(Type.Graph)) {
-                graphs.push([name, value as ClassMetaType]);
-            }
+    private getProgramContext(plugin: Plugin) {
+        let contextPromise = this.programs.get(plugin);
+        if (contextPromise === undefined) {
```
slaymaker1907: Change this to just be if(contextPromise) to avoid weird issues with null vs. undefined.

plugins/dfa-interpreter.ts
```diff
@@ -1,207 +1,102 @@
...
-        this.nodes = serialGraph.nodes.map((oldNode: any) => {
-            const result: Node = {
-                label: oldNode.drawableProperties.Label,
-                parents: [],
-                children: []
-            };
-            transferProperties(oldNode, result);
-            return result;
-        });
+export type Nodes = DFANode
```
slaymaker1907: I think these should end with semicolons.

plugins/dfa-interpreter.ts
```diff
@@ -1,207 +1,102 @@
...
-}
-
-interface PropertyObject {
-    [propName: string]: any;
-}
-
-interface Element extends PropertyObject {
+export class DFANode {
+    isStartState: boolean;
+    /** Accept State */
```
slaymaker1907: It doesn't seem like we need this renamed?

plugins/dfa-interpreter.ts
```diff
@@ -1,207 +1,102 @@
...
+
+        if (node.isAcceptState) {
+            accepts.add(node);
         }
-        let symbols = node.children.map((edge: Edge) => edge.label);
-        let uniqueSymbols = new Set(symbols);
-        if (symbols.length !== uniqueSymbols.size) {
-            throw "This interpreter does not handle NFAs. Non-unique edge label detected.";
+
+        if (node.children) {
```
slaymaker1907: children should never be null and should just be an empty list if there are no children.

app/components/graph-editor/README.md
```diff
@@ -1,6 +1,98 @@
-_Last updated: January 31, 2017_
-
-# Notes
+_Last updated: February 11, 2017_
+
+# Remarks
+The graph editor component is a custom Angular2 component that is intended to be
```
ilargierdi: Maybe in the future, but for now, this is what it is.

app/components/graph-editor/drawable-element.ts
ilargierdi: I'm not sure how that would be profoundly different? Other than having an extra type declared somewhere.

app/components/graph-editor/defaults.ts
ilargierdi: I agree. Currently this entire file is a stand-in for what should be a global user settings file, except that it currently only contains stuff related to the graph editor.

app/models/main-graph.ts
Sheyne: What?

app/models/main-graph.ts
Sheyne: It would be very complicated.

app/components/graph-editor/events.ts
ilargierdi: Is there a compelling reason why it should be an interface instead of a class?

app/models/main-graph.ts
Sheyne: But Daniel needs the details of the change. You could make it also conform to the interface, but this is more data driven rather than encoding the logic of undoing the change into the event.

app/models/main-graph.ts
Sheyne: Readonly.

app/components/graph-editor/README.md
RighteousRaichu: I think the sentiment is it should be decoupled such that we could move it to its own package (and it is). But that currently moving it wouldn't provide a tangible benefit towards completing Beta.

app/components/graph-editor/drawable-element.ts
RighteousRaichu: I think we should focus on readability. Short imports shouldn't need two lines, long imports can be too long if we don't break them up.

app/models/main-graph.ts
RighteousRaichu: I agree with Sheyne. The changes are useful beyond Undo/Redo. The current implementation is the most workable IMO.

app/services/plugin.service.ts
RighteousRaichu: I agree. In general, I prefer `if (value)` checks for checking if a value is defined and explicitly checking for booleans. (ie `if (value == true)`)

plugins/dfa-interpreter.ts
RighteousRaichu: @Sheyne why is the naming like this? Did you just wanna try out both ways or having a pretty name?

plugins/dfa-interpreter.ts
RighteousRaichu: I made lots of changes to the interpreter just to make it work more as expected. In the general case we have #107 to make sure the values sent to the interpreter match what it's expecting.

app/components/graph-editor/defaults.ts
RighteousRaichu: This probably won't happen until after Beta, but I imagine the settings window could have some behavior where each component that wants settings can have their say. So the settings window would have any number of sections, and the IDE could have what it wants, the graph editor could have what it wants, the plugin service could have what it wants, etc...

app/components/properties-panel/properties-panel.component.html
```diff
@@ -1,77 +1,41 @@
...
-                    <td class="label">{{pair[0]}}</td>
-                    <td [ngSwitch]="pair[1].kind">
-                        <div *ngSwitchCase="'Number'">
-                            <input [ngModel]="selectedEntity[group[1]].get(pair[0])" (ngModelChange)="doChange(selectedEntity, group[1], pair[0], $event)" type="number" 
-                                class="number-input"/>
+    <div *ngIf="!isEmpty">
+        <table *ngFor="let name of fieldNames" class="group-table">
+            <template [ngIf]="fields[name]">
+                <tr><th colspan="2">{{name}}</th></tr>
+                <tr *ngFor="let triple of fields[name]">
```
RighteousRaichu: Unless you have a better idea, @Sheyne, I'm going to make a function that wraps up triple in a object with named properties (ie `let triple of wrap(fields[name])` -> `triple.name`, `triple.path`, etc...) for readabilities sake.

app/components/properties-panel/properties-panel.component.ts
```diff
@@ -3,90 +3,49 @@
...
-
-    private isEmpty() {
-        if (this.selectedEntity) {
-            for (let group of this.groups) {
-                let g = this.selectedEntity[group[1]];
-                if (g.properties.length > 0) {
-                    return false;
-                }
-            }
+    isEmpty = true;
```
RighteousRaichu: We should be more careful in general about marking stuff as private.

app/components/graph-editor/README.md
slaymaker1907: Sounds good. I mostly just wanted to know what the overall plan for that was.

app/components/graph-editor/defaults.ts
slaymaker1907: I'm also planning to support plugin specific settings as well. Not sure exactly how that will work with the overall plugin API, though.

app/components/graph-editor/drawable-element.ts
slaymaker1907: I just think that abstract classes usually indicate that there is coupled functionality (since some is provided by the abstract part and some by the subclass) that should be decoupled through composition.

app/components/graph-editor/events.ts
slaymaker1907: Interfaces usually have less coupling and are easier to extend (i.e. don't have to correspond to inheritance hierarchy).

app/models/main-graph.ts
slaymaker1907: Something like

```
interface Mappable {
  [a: string]: any;
}
```

app/models/main-graph.ts
slaymaker1907: Sounds good

app/services/plugin.service.ts
slaymaker1907: According to what I read about Angular depedency injection, even though the @Inject is a specific type, we can change it to be something else that fulfills the FileService interface.

plugins/dfa-interpreter.ts
slaymaker1907: I don't think that this is related to #107. With the version of the serializer that I wrote, children was explicitly set and was never null. I'm not sure if it was changed so that this could be null or if it is an unnecessary check.

RighteousRaichu:
* DONE Rename to GraphController
* DONE Make sure to specify which plugin kind in the error message.
* DONE Rename any in Program declaration
* DISCUSSED Change this to just be if(contextPromise) to avoid weird issues with null vs. undefined.
* DONE I think these should end with semicolons.
* FUTURE Remove triple[…] in properties-panel.component.html <- Will be done when using custom components to render these.
* DONE Serialize/Deserialize errors through the sandbox, use this to rehookup the status bar at least for compiler errors.
* DONE Mark fields in properties panel component as private

PR #144, opened 2017-02-19
Light theme & SASS

RighteousRaichu:
* Made left panel wider and taller by moving icons to the top, making the bottom panel less wide. (This makes each panel a more useful size)
* Added `ng-inline-svg` as a dependency so our svgs can be styled with CSS.
* Consistently formatted (S)CSS. 
* Removed debug panel. Debugging functionality will be part of the input panel.
* Made the unit test panel slightly spiffier, but there’s still a lot of work to do in this panel.
* Added SASS to the build pipeline and converted most CSS files to SCSS. (This got rid of the long standing TODO in index.html, woo!)


RighteousRaichu:
A lot of CSS changes, so if you wanna review cool but otherwise just build it and make sure you don't see anything too ugly.

PR #145, opened 2017-02-19
Use `electron-renderer` webpack target

RighteousRaichu:
...which fixes a lot of electron specific problems we've had with webpack... wish I would've known this target existed earlier haha.

(Sorry about the weird commit history.)


RighteousRaichu:
This isn't a big change and is blocking my other PR. So I'm going to merge this soon unless someone really wants to look at it.

Sheyne:
Go for it.

PR #146, opened 2017-02-19
Shortcut for closing tabs and switching tabs

RighteousRaichu:
Can you guys make sure this works well on your respective OSs?

(Sorry about the weird commit history.)


ilargierdi:
If no tabs are open, and we do Ctrl + W, should the window close?

RighteousRaichu:
Yeah @ilargierdi, is that not working?
(Actually, you should mess around with some apps on your OS and see if that's how they behave.)

ilargierdi:
I tried it on a couple browsers. Firefox and Chrome do close their windows, but Opera just keeps the one default tab open. I think I'd rather have the window stay open even if you keep spamming `Ctrl + W`. All the other shortcuts are working nicely, though.

RighteousRaichu:
So currently on not-macOS it DOES close the window if there are no tabs, but you want it to NOT do that, correct?

ilargierdi:
Correct.

Sheyne:
For macOS we are making it close the app?

RighteousRaichu:
Yeah I'll make it close the app only on macOS (if there are no tabs.)

PR #160, opened 2017-02-22
Use angular-resizable-element to resize panels

RighteousRaichu:
There's a directive that will do this for us. So for now we'll use that.


RighteousRaichu:
Okay it now limits to a maximum size for the panels, in a way that's good enough for me for beta. A future issue is making sure when the window resizes, etc... that nothing looks glitchy.

PR #165, opened 2017-02-27
Graph editor images

ilargierdi:
* Added support for custom images
  - `Shapes` now has an additional type in its union: "image"
* Added support for custom anchor points
* Drawable nodes now have two new public fields: `image` and `anchorPoints`
  - The `image` field is a string representing the path (preferably absolute) to the image file
  - The `anchorPoints` field is a list of points
    + Points with duplicate coordinates are removed
    + The coordinates of an anchor point is relative to the upper left corner of a node

### DIFF
--- a/app/components/graph-editor/defaults.ts
+++ b/app/components/graph-editor/defaults.ts
@@ -104,6 +104,8 @@ export const NODE_PROPERTIES = {
     get position() { return { get x() { return 0; }, get y() { return 0; } }; },
     get label() { return ""; },
     get shape() { return "circle"; },
+    get image() { return ""; },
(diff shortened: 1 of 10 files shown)


ilargierdi:
Yes, I can do that tonight. Although, I think for the accept states, I'd rather just add in a new double line style for the border.

RighteousRaichu:
Does "double" borderStyle work on nodes yet?

ilargierdi:
Not yet

RighteousRaichu:
Do you want to do it for this branch or should I just merge this?

ilargierdi:
Just merge, and I'll make a separate issue for double lines.

Issue #134, opened 2017-02-18
Global support for undo/redo


Sheyne:
Should intelligently include the graph editor, interleaving undo events from the OS with `Node` moves and whatnot.

RighteousRaichu:
Working on this in: https://github.com/2graphic/sinap-ide/tree/undo-redo

RighteousRaichu:
Some comments here: 422932d

RighteousRaichu:
@CJ would it break anything to merge the listeners for creating and deleting nodes and edges, so we'd just have one listener for creating and deleting drawable elements?

Right now if say I delete a graph that contains nodes and edges, I'll get a change for the deleted nodes and a change for the deleted edges, and the graph controller has no way of knowing if that was one action or not for undoing.

RighteousRaichu:
Also moving an edge still triggers two notifications instead of lumping them together.

RighteousRaichu:
FYI @ilargierdi I'm working on this branch this afternoon

ilargierdi:
I'm done working on it, unless there's something horribly wrong with my
changes that you need me to fix.

RighteousRaichu:
@CJ I need you to look into a couple things:
* DrawableGraph keeps a list of _deleted elements, which is unacceptable because this can grow without bound.
* When you create two edges between nodes, such that they're curved, and delete one, the other one stays curved (this has been happening since your previous commit.)
* I had to do some hacky things to DrawableGraph to make this work, you may wanna change some stuff. IMO we may wanna rethink the events in 1.0 and leave it hacky for now.

RighteousRaichu:
After beta we need to discuss a more powerful event infrastructure in order to clean up this code.
Also need to support undo'ing moving elements.

PR #170, opened 2017-02-28
Undo & Redo

RighteousRaichu:
It's finally almost ready, just a couple points still need to be addressed in #134

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,10 @@ before_install:
 install:
 - npm install
 script:
+- npm run lint
 - npm run build:dll
 - npm run build
 - npm run build:prod
-- git diff --quiet HEAD
 notifications:
   slack:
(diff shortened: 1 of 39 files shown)


ilargierdi:
I'm currently working on this, and I think I have a solution that'll work
better than what we have now.

RighteousRaichu:
One thing I forgot is being able to undo moves. Preferable for each drag and drop event one notification is sent to graph controller containing all the elements that were moved as well as old/new location. Whatever you're refactoring see if you can work that in.

ilargierdi:
@RighteousRaichu See if this works for you b5f2681.

RighteousRaichu:
I think I'm going to revert back to this morning and merge that. After beta we need to rethink the way notifications works, but currently they aren't powerful enough to accomplish what we need to get done.

(For instance I think you forgot we need to be able to delete/undelete groups of elements (both edges and nodes), as well as undo moving groups of elements.) (And eventually undo changing properties of groups of elements but that's not really the graph editor's concern.)

Also just by creating nodes, edges, moving around nodes and edges, and undo'ing + redo'ing I can get the graph out of sync pretty easily with these changes.

ilargierdi:
Drawable events do have support for what you're talking about. You just
don't get them from the controller, but you can; the undoable events just
need to be augmented for that. As for copying properties to the core
elements, that can also be easily done in the controller silently with the
changes I have in place.

RighteousRaichu:
So currently when a group of elements are deleted, two events go out, one for nodes and one for edges. These need to be one cohesive unit to be undoable. I can't see a way to do it with the way notifications are currently setup in drawable-graph without essentially doing what I've already done in previous commits.

There's three main regressions here:
* Can't delete/undelete groups of elements.
* Plugin properties aren't retained for undeleted elements.
* Playing around with undo/redo even without deleting groups of elements eventually causes a bad state (edge with no nodes attached, etc...)

I've looked into this problem long enough that I have a good idea for how I want to change the way notifications work in order to remove the 100's of lines of buggy/will become buggy/unnecessary code from graph-controller and drawable-graph so let's just talk about it again after beta.

PR #176, opened 2017-03-02
Plugin Loading

RighteousRaichu:
@slaymaker1907 here's the PR for your stuff. Make sure everything's in here and still works.

### DIFF
--- a/app/services/files.service.ts
+++ b/app/services/files.service.ts
@@ -49,7 +49,8 @@ function ensureNull(shouldBeNull: any): Promise<void> {
 export class LocalFileService implements FileService {
     getAppLocations(): Promise<AppLocations> {
         const currentDirectory = new LocalDirectory(app.getAppPath());
-        const pluginDirectory = new LocalDirectory(path.join(app.getPath("userData"), 'plugins'));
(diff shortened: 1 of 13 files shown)


app/services/files.service.ts
```diff
@@ -49,7 +49,8 @@ function ensureNull(shouldBeNull: any): Promise<void> {
 export class LocalFileService implements FileService {
     getAppLocations(): Promise<AppLocations> {
         const currentDirectory = new LocalDirectory(app.getAppPath());
-        const pluginDirectory = new LocalDirectory(path.join(app.getPath("userData"), 'plugins'));
+        const pluginPath = process.env.ENV !== 'production' ? path.join('.', 'plugins') : path.join(app.getPath("userData"), 'plugins');
```
slaymaker1907: This seems to work ok, at least for now. However, make sure to keep the code for copying over the directory since we will likely want to have a version of the app that comes with preinstalled plugins as suggested to me by @RighteousRaichu

RighteousRaichu: I think what you were talking about with needing an installer might be necessary for windows as we can't really copy files out of the .exe

PR #181, opened 2017-03-04
Stacked edge labels

ilargierdi:
Hit detection also works on the floating labels for edges. Depending on which horizontal half of the label the cursor is on, either the source or destination endpoint of the edge will be picked up when dragging.

The labels will always be on the same side of the edge, regardless of the edge's global orientation. To see this in action, create two nodes and four edges with labels between them--each pair of edges opposing the other--and orbit one of the nodes around the other by dragging.


Sheyne:
Cool

ilargierdi:
I forgot to mention that the edges will be stacked according to their natural ordering.

Issue #171, opened 2017-02-28
Node recenters when picked up


Sheyne:
Clicking to select node also recenters :(

ilargierdi:
#183

PR #183, opened 2017-03-04
Address #171

ilargierdi:
* Clicking or dragging a node no longer recenters the node to the cursor.
* The primary node being dragged is now drawn on top of everything.

Feel free to squash.


RighteousRaichu:
screen shot 2017-03-04 at 10 30 22 am

This is really great! But I managed to somehow do what you see in the screenshot. I think it involved moving edges between nodes but I'm not sure. If you can't reproduce I can mess around with it more.

ilargierdi:
Can you tell me what you did to get that?

RighteousRaichu:
Ok I played around and figured it out. If you create two nodes a & b, then an edge from node a->b, then drag the edge destination as if you were going to move it to a new node and just let go (so the edge doesn't get moved.), then move node a.

ilargierdi:
Try it now.

Issue #177, opened 2017-03-02
Graph Editor should remember zoom and pan when switching tabs


RighteousRaichu:

RighteousRaichu:
This is done when switching tabs. However, it should probably also be remembered when saving/loading files as well. (Possibly hidden readonly properties on a graph?) Leaving this as 1.0

ilargierdi:
I believe this works from #215 now.

PR #186, opened 2017-03-06
bugfix(#177)

ilargierdi:
* Drawable graphs now keep track of pan and zoom.
* Anchor points are now properly bound when creating an edge.


ilargierdi:
Also, I'm unable to replicate that horizontal edge bug on Windows again, so I have no idea what's going on with that.

RighteousRaichu:
I also can't reproduce the bug currently, we'll have to be on the lookout for it. Everything here looks great though!

PR #195, opened 2017-03-06
Added Descriptions to Plugins

slaymaker1907:
It will require some changes to support fully nested plugins, but this seems reasonable for now.


package.json
```diff
@@ -9,6 +9,7 @@
   },
   "main": "index.js",
   "scripts": {
+    "install": "npm run build:dll",
```
RighteousRaichu: I disagree, the whole point of build:dll is that you shouldn't need to do it ofter.

ilargierdi: I think we only want to run `build:dll when` we upgrade to a new version of angular, is that right? That, or if we change any of the packages that are part of our `vendors` module in webpack.

slaymaker1907: I can remove it, but I thought it made sense to do whenever we do an npm install since you theoretically run install infrequently and that would be when you upgrade Angular and what not.

It also makes a more conventional build process in my opinion since people only need to run

```
npm install
npm run build
```

Sheyne: Anyone mind if I move this out of this PR and into it's own?

RighteousRaichu: Yeah do that Sheyne and we can discuss this IRL eventually

app/components/new-file/new-file.component.ts
```diff
@@ -21,58 +22,46 @@ export class NewFileResult {
 })
 export class NewFileComponent implements ModalComponent, AfterViewInit {
     set modalInfo(modalInfo: ModalInfo) {
-        let plugins: string[][] = modalInfo.data;
-        let availablePlugins: any = { subLists: {} };
+        const plugins: PluginData[] = modalInfo.data;
+        // This code is extremely evil but necessary.
```
Sheyne: Wait, what's going on here?

Sheyne: Couldn't this evil at least happen in the plugin service? Or better yet, why can't this just be done without the evil in core?

slaymaker1907: That is a fair point that I agree with. However, I don't think that I have time now to fix it. This would be a good issue to tackle when I refactor core to do descriptions according to npm. What I think would be ideal is to have a proper serialization mechanism in core for plugins.

Sheyne: What is it doing? It seems like its just adding a function to the plugin object rather that adding that function in core. If so it's a trivial change to put it in the right place. I may be missing something though.

slaymaker1907: The reason it is necessary is because they need toString to work properly with the collapsible list component. However, prototype information gets stripped out when something is serialized so I just add it back in. It's probably still unsafe, though it is safer than it first appears.

Sheyne: But the plugin doesn't get serialized. What's going on?

slaymaker1907: Also, I would like to avoid instantiating the plugin service in this window since that will make the load time of the window even worse than it already is until we cache the window.

Sheyne: Oh. That makes sense, serialized to make it to the window. okay.

PR #196, opened 2017-03-06
Added build:dll to install script.

Sheyne:
As requested by @slaymaker1907


RighteousRaichu:
What if we hook `npm update` instead?

Sheyne:
That still means you have to build: dll on your initial install

slaymaker1907:
Can we merge this soon or close the PR?

ilargierdi:
I'm closing this, since I think the majority of us don't feel there is any benefit to the proposal.

PR #199, opened 2017-03-06
fix validate messages

RighteousRaichu:


RighteousRaichu:
@sheyne do you like that better?

PR #209, opened 2017-03-07
Circuits

RighteousRaichu:
Make sure nothing is borken pls


ilargierdi:
How do I run the input?

Sheyne:
Check or uncheck a box the hit enter.

ilargierdi:
Hitting enter does nothing

Sheyne:
Hmm at least on Mac if you give a check box focus enter works. It's wonky but probably good enough for demo. Eventually we probably want shift enter to run the input regardless of selection.

ilargierdi:
Oh, the checkbox needs to have the focus; it works, but only if either of the check boxes have focus.

ilargierdi:
Is the turing machine supposed to output a boolean in the input panel?

RighteousRaichu:
Yeah, Turing Machines output whether or not they accept the language.

ilargierdi:
Shouldn't it always return true? I think it'd be neat to show the tape as part of the output, but it's not that important for the presentation.

Sheyne:
Formally Turing machines decide if a string is in a language. The tape is "irrelevant". For example deciding whether a number is divisible by 5, or is prime, or whatnot. Most of the time you care more about the acceptance than the tape content.

RighteousRaichu:
I'm planning on saying something around those lines, and saying, "we can step through execution and see how the tape changes with the debugger."

PR #215, opened 2017-03-17
Graph editor maintenance

ilargierdi:
Everything should be working as it did during our beta presentation. Let me know if you catch any bugs related to the editor or graph controller.

## Changelog
* Changed drawable events API
  - Now uses `CustomEvent` and `EventTarget` to look more like DOM events
* Moved draw, update, and hit detection logic out of drawables
  - Drawables now only contain properties and a few methods
* Added move event to `DrawableGraph`
* Added recreate method to `DrawableGraph`
* Changed graph editor drawing behavior by performing a graph traversal
* Changed graph controller undoable events
  - This might get scrapped, and also needs more thorough testing
  - Some basic tests were performed
    + Element creation/deletion, edge moving, and node position change
* Some fields have been made enumerable on drawables
  - The graph now exposes the origin and scale for file saving
  - Edges expose source and destination points for saving anchor points
  - Nodes expose an origin and a size
* Added more thorough API documentation
  - Using some of the tags from JSDoc


RighteousRaichu:
I think maybe custom images no longer ever look selected.

RighteousRaichu:
I can also get errors by undoing and redoing, but rather than debug it now, I think I'm going to ignore it because graph controller has changed so much in the CoreValue branch.

RighteousRaichu:
Also +1 for making squares fix the size of the label instead of always being a square.

Issue #167, opened 2017-02-27
Get path to plugin directory


RighteousRaichu:
Each loaded plugin should probably contain the path to its folder. This would be helpful for being able to reference images by just a relative URL.

slaymaker1907:
What do you mean exactly? Plugins are already directory based with package.json.

Sheyne:
He means that given some plugin object, he wants to know what directory it came from so he can load assets from that directory.

RighteousRaichu:
Like Sheyne said, a plugins assets should be inside the plugin's folder. And the plugin should be able to reference images/etc by a relative path. The first step is having all plugin objects have a path to their directory.

RighteousRaichu:
@slaymaker1907 is this done?

slaymaker1907:
This is available through InterpreterInfo which is associated with a Plugin.

PR #216, opened 2017-03-18
Mostly Fixed Packaging

slaymaker1907:
There are still some mysterious bugs with some of the plugins where you can't create edges for some reason (the circuit plugin seems to work, though).

A couple of changes I also implemented was to give process.env.ENV a less confusing name of sinap.env. I also removed the boundary for the new file window which allows for developer tools and a clean menu.


slaymaker1907:
See #217 for the remaining issue.

ilargierdi:
I can't open anything on a fresh clone on CentOS. The console is saying that it can't find the plugin associated with the file. I tried going through the new file dialog box, but nothing would load there, either.

slaymaker1907:
@ilargierdi make sure that you are using npm version >= 4.2 (the lowest I've gotten to work). I was able to reproduce your error on Ubuntu, but it turned out I was using npm 3 instead of npm 4. Once I updated npm via

```bash
sudo npm install -g npm
````

it now seems to work. I also added a note in the README.md about this dependency.

**Comment by ilargierdi** [comment]
I'm still getting the error. I'm running on npm 4.4.1 and node 6.9.4.

**Comment by slaymaker1907** [comment]
I will install a CentOS VM and see if I can replicate this.

**Comment by slaymaker1907** [comment]
I was able to get everything working on a CentOS 7 VM. I can send you a copy of the VM if you would like to try it out and compare it to your setup. For some reason it wouldn't build Windows binaries, but I think that may just have been particular to my setup and not Sinap's fault.

**Comment by ilargierdi** [comment]
If you're talking about running from the dist folder after `npm package`, that works fine. I'm talking about `npm start`; I can't open anything or even create a new file if running from the build directory. Also, I'm not sure if I like the borderless dialog window. Mostly because I miss the close button at the top.

**Comment by slaymaker1907** [comment]
@ilargierdi I think your issue was that you did not run a regular build after running package. The production build is slightly different than the development build, in particular regarding where plugins are loaded.

I just pushed a small change so that it will remove the build folder after packaging to try and help avoid these issues.

build:dll is the same in both production and in development so I don't delete that folder.

**Comment by ilargierdi** [comment]
The bug I'm experiencing was happening before I event attempted to `npm run package:linux`. I hadn't even used that command until my last comment. From a fresh install, build:dll, and build, if I then try `npm start`, the application runs fine, but I can't open anything or even create a new file.

**Comment by slaymaker1907** [comment]
Ok, I think I fixed the issue. I misunderstood the electron app.md documentation.

**Comment by ilargierdi** [comment]
It works now!

**Comment by slaymaker1907** [comment]
I implemented the suggestion of @RighteousRaichu to make the new file window bordered as well as adding in a debug flag separate from the production flag.

**Comment by slaymaker1907** [comment]
@Sheyne or @RighteousRaichu can you try this out on MacOS?

File: app/models/menu.ts [hunk 1]
```diff
@@ -242,16 +242,11 @@ if (process.platform === 'darwin') {
...
- * Only allow the Developer Tools to be triggered in a development build.
- */
-if (process.env.ENV === 'development') {
-    let viewSubmenu = viewMenu.submenu;
-    if (viewSubmenu && viewSubmenu instanceof Array) {
-        viewSubmenu.push({
-            role: 'toggledevtools',
-        });
-    }
+let viewSubmenu = viewMenu.submenu;
```
RighteousRaichu: I'm gonna add a check here for debug mode. Any objections @slaymaker1907

RighteousRaichu: `(node:75895) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: ENOENT: no such file or directory, open '/Users/Danie
l/Development/sinap-ide/build/package.json'`

I'm gonna look into this but any ideas @slaymaker1907 ?

RighteousRaichu: I think we should address #167 before merging this in. Currently from a packaged build I can't open examples that have a custom image because it can't find the image file.

RighteousRaichu:
Some things we need before we can merge this:
Figure out why on my computer I can't package twice in a row without cleaning first.
Find workaround for missing typescript definition files.
Expose plugin directory to Plugin so custom images work.

slaymaker1907:
I haven't figured out what's been going on for @RighteousRaichu with duplicate package. However, typescript lib files are now bundled in the app and the plugin directory is exposed to the graph editor (kind of a hack for now until we do plugin specific resources which requires stuff from core).

With these changes, we should be able to ship the app to third party users.

RighteousRaichu:
I can't reproduce the problem anymore and everything looks great rn. Feel free to merge this in when you're ready @slaymaker1907

PR #229, opened 2017-03-25
Image selection highlighting

ilargierdi:
Reintroduces image highlighting for selected nodes with custom images.


RighteousRaichu:
There's a bug that was from your last pull request I just noticed. When states are highlighted by the input panel, they don't show as highlighted until there's a mouse movement in the graph editor.

PR #245, opened 2017-03-29
New CoreValue Changes

RighteousRaichu:
I'm waiting until Friday for some changes from Sheyne, so I'm thinking I'll merge this into master soon. Needs some cleanup work first.


ilargierdi:
UI improvements are looking great! I dig the new look of the properties panel. The text on the left side of the blue bar needs a little bit of padding, since it was showing up against the left edge.

I'm not sure how much you're aware of being broken, so I'll just list what I noticed:
* position doesn't update in the graph editor when changed in the properties panel
* adder file doesn't load
* input panel wasn't working; I suspect because you're waiting for the changes from Sheyne
  - I tried testing it on the NFA and DFA, and I imagine the unit test panel would have similar results
* there's a fair bit of lag with live updating the graph editor when rapidly changing the label of a node
  - this may be an issue with the graph editor and unrelated to your changes

Unrelated to your changes, I noticed that the boundary anchor point is laggy for large nodes and boundary point calculations are incorrect for ellipses. This is more of a mental note for me to investigate when I have time. I also created #246 to reference later when I get around to it.

Sheyne:
I'd say that the fact that `node package.js` produces executables that don't run is the only thing blocking merging this. example

Sheyne:
So the first part of the above issue is that the build dir isn't getting copied to the application bundle anymore. This is we're because package.js hasn't really changed. If I manually copy the build dir, it also fails, but it at least shows the main GUI (console complains about missing "./plugins" dir).

RighteousRaichu:
I'm looking into it now.

package.json
```diff
@@ -7,7 +7,7 @@
     "type": "git",
     "url": "https://github.com/2graphic/sinap-ide"
   },
-  "main": "index.js",
+  "main": "build/index.js",
```
RighteousRaichu: @Sheyne is this change necessary?

slaymaker1907:
This is like -3000 LoC. I approve.

Issue #264, opened 2017-04-07
Border styles for nodes carry over to label and selection shadow


ilargierdi:
Title.

ilargierdi:
#267

Issue #223, opened 2017-03-24
Update graph editor input controls


RighteousRaichu:
Should we make a double click in the graph editor add a node?
The reasoning is that when I want to make many nodes all at once, the current behavior is very slow.

This could in addition to or replace the current behavior.

I believe we also initially talked about a single click to add a node, and click and hold to pan the graph?

Sheyne:
I like for macOS pan and zoom with two fingers and single click makes a node.

ilargierdi:
From our meeting on 3/24:
```
* Double click to create node
* Select multiple by dragging with shift
  - also by pressing down, not pausing, then dragging
* Move canvas by dragging with option (alt)
  - also by pressing down, pausing, then dragging
```

ilargierdi:
#267

PR #267, opened 2017-04-09
Editor input changes

ilargierdi:
Addresses #264 and #223 

There's a bug with double clicking that I couldn't figure out. Every time the user double clicks on the canvas, the text in the status bar is selected regardless of whether or not `stopPropagation`, `stopImmediatePropagation`, or `preventDefault` is called on the event payload.

### DIFF
--- a/app/components/graph-editor/drawable-graph.ts
+++ b/app/components/graph-editor/drawable-graph.ts
@@ -570,19 +570,17 @@ export class DrawableGraph extends Drawable {
     ) {
         const oldSelection = [...this._selected];
         move(src, dst, items, v => v.isSelected = (dst === this._selected));
-        if (oldSelection.length !== this._selected.size) {
-            this.dispatchEvent(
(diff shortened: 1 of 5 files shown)


RighteousRaichu:
@cj see if my commit also fixes it on Windows/Linux and feel free to move that code somewhere better.

And I definitely like the input scheme better.

Some things:
* When a node is created it should be selected.
* When the graph is clicked it should be selected.
* When a node has a dotted outline, and is hovered over (and not selected), you can see solid blue behind the dotted outline.

RighteousRaichu:
I guess for

> When a node has a dotted outline, and is hovered over (and not selected), you can see solid blue behind the dotted outline.

It has done that this whole time.

ilargierdi:
> When the graph is clicked it should be selected.
The editor correctly deselects everything when the user clicks on empty space, so I'm not sure why the properties panel doesn't show the graph as being selected.

PR #276, opened 2017-04-13
Faster modal window

RighteousRaichu:
@cj can you test this on Windows and Linux when you get a chance?
Mainly test to make sure lifecycle stuff hasn't broken. (Can still open new window dialog, can still close app, etc...)

### DIFF
--- a/app/components/new-file/new-file.component.html
+++ b/app/components/new-file/new-file.component.html
@@ -11,10 +11,10 @@
         </div>
         <div id="filename" class="flex-container-row">
             <div id="text-input">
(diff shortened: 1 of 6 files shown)


ilargierdi:
image
Menu bar shouldn't be there on Windows; it's also showing up on Linux. Aside from that, it works great!

slaymaker1907:
@ilargierdi is that from the packaged version or from the development version? The menu bar must be shown on Windows if debug is enabled so that you can open up developer tools.

ilargierdi:
Oh right. It was on the development version.

ilargierdi:
Nevermind, it's also showing up on the packaged version for windows.
image

PR #278, opened 2017-04-13
Plugin import

slaymaker1907:
Added a really basic looking window for plugin management. It is just a couple of buttons at the moment, but the backend work is all there to support managing plugins individually.

I also added in a slightly more sophisticated logging mechanism so that I can have more informative logs and don't have to constantly add and remove console.logs. Feel free to use it yourself if you find it useful.


slaymaker1907:
Would @RighteousRaichu or @Sheyne test these features on macOS with a packaged version? I inherently have to modify the plugins directory and this might cause similar issues to what we saw earlier with packaging for macOS.

Sheyne:
I'm not sure if I'm doing it right, but I imported a plugin and it didn't show up in the new file dialog.

Sheyne:
So it works, but I have to reload the IDE to actually use the plugin.

slaymaker1907:
This should now use appData when in production mode (by default it just creates the directory and does not import plugins by default).

slaymaker1907:
Looks pretty, though hot reloading of plugins is still somewhat buggy for some reason. In order to work properly, you need to reload.

PR #311, opened 2017-04-20
Added change watching plus a promise lock.

slaymaker1907:
It is still a bit slow unfortunately when managing plugins because it forces a full reload.

The lock thing is necessary versus just a regular reload since there are multiple loading periods.


RighteousRaichu:
@dyllon deleting plugins doesn't work as expected, lots of errors get thrown from the watching.

PR #319, opened 2017-04-21
more portable access to VSCode

Sheyne:


Sheyne:
I mean, it's POSIX now at least.

slaymaker1907:
Would it be better to use this https://electron.atom.io/docs/api/shell/ to open it up? You would have to have .ts files default to Code for opening, but it would be extremely portable.

Sheyne:
@slaymaker1907 I'd really like it if we could find a way to make it open folders in text editors such as VSCode, Sublime, or Notepad++

PR #321, opened 2017-04-21
Added ability to open a plugin in the file manager.

slaymaker1907:
I think this is nice for more general/complicated plugins. I'm going to work on the duplicate folder issue, but this was a quick fix.

This should be cross platform since I'm just using a built in electron API.

@Sheyne I'm really interested as to what you think.

Also fixed a bug with getPath for Windows.

### DIFF
--- a/app/util.ts
+++ b/app/util.ts
@@ -57,7 +57,7 @@ export function somePromises<T>(promises: Iterable<Promise<T>>, logger: Logger):
 }
 
 export function getPath(name: string) {
-    return path.normalize("/" + path.relative("/", name));
+    return path.resolve(path.normalize(name));
 }
 
 export function compareFiles(file1: string, file2: string) {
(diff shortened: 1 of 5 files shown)


app/util.ts
```diff
@@ -57,7 +57,7 @@ export function somePromises<T>(promises: Iterable<Promise<T>>, logger: Logger):
 }
 
 export function getPath(name: string) {
-    return path.normalize("/" + path.relative("/", name));
+    return path.resolve(path.normalize(name));
```
RighteousRaichu: Does this still return the full path?

slaymaker1907: It should, and the older version did not work properly on Windows.

slaymaker1907: Somebody should probably test on Mac/Linux.

PR #322, opened 2017-04-21
Undo/Redo v2

RighteousRaichu:
#134 

~There's still the bug with adding back ElementValue's to the Model. Occasionally you can trigger a "Value not in Environment" error, still.~ #326

Other than that hopefully this doesn't cause errors.

I've debounced undoable events to ~200~ 100 milliseconds. This handles cases such as when you move multiple nodes.. or in the future when the properties panel can edit multiple nodes at a time.

I'll probably keep this regardless, as it allows me to debounce how often a new Program is compiled from the plugin, though events to the graph controller should still be as debounced as possible. 

I'm also trying to make a distinction between events that are merely visible, and events that should trigger the plugin to recompile. (moving nodes doesn't trigger a recompile, when I add zooming, nor will that.)

Note: origin and zoom and currently not saved to the core model and therefore not undoable... working on it.

### DIFF
--- a/app/components/graph-editor/drawable-graph.ts
+++ b/app/components/graph-editor/drawable-graph.ts
@@ -125,7 +125,10 @@ export class DrawableGraph extends Drawable {
                 enumerable: true,
                 get: () => this._origin,
                 set: (value: point) => {
-                    const old = this.origin;
+                    const old = {
+                        x: this._origin.x,
+                        y: this._origin.y
+                    };
                     if (value.x !== old.x || value.y !== old.y) {
                         this._origin.x = value.x;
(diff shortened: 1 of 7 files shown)


RighteousRaichu:
@sheyne or @cj, I think is ready to be merged in, can someone give it on more look through?

ilargierdi:
Mostly works on Windows. The only issue is that for property changes, I have to undo three times and redo three times to see the changes. Also, I'm not sure if this was a problem previously with Windows, but the keybindings for undo/redo are working, now.

RighteousRaichu:
@cj which properties are you playing with?

PR #323, opened 2017-04-22
multitouch pinch/zoom for macOS

Sheyne:
This will mess with the current behavior for windows. I think it might be good, but I'm happy to special case macOS. In general I think panning is more common than zooming, so having it pan unless the ctrl key is pressed makes sense to me.


ilargierdi:
+ Bonus random bugfixes

PR #345, opened 2017-04-24
Made sure to ensure plugin dir is created.

slaymaker1907:
The main issue was that I was not making sure that a plugin directory was created if none already existed.

Also improved performance of managing plugins by pruning duplicate acquisitions of the lock.


slaymaker1907:
@Sheyne I've greatly improved the performance of the manager by making the lock eliminate duplicate events as much as possible.

slaymaker1907:
I'm going to take a break from IDE now to work on the cli.

PR #358, opened 2017-05-04
Made validateEdge check type.

slaymaker1907:
Also added sinap-core as explicit dependency since we need 3.11.4 or higher. Note that we now create a fake edge for use with validateEdge.

### DIFF
--- a/app/models/graph-controller.ts
+++ b/app/models/graph-controller.ts
@@ -80,7 +80,6 @@ export class GraphController {
             destination = destinationBridge.core;
         }
 
-        // TODO: ...like
         let likeVal: ElementValue | undefined = undefined;
         if (like) {
             const bridge = this.bridges.getB(like);
@@ -89,11 +88,19 @@ export class GraphController {
             }
         }
 
+        let cleanup = false;
         if (!likeVal) {
+            cleanup = true;
             likeVal = new ElementValue(this.activeEdgeType, this.core.environment);
         }
 
-        return validateEdge(this.plugin, source, destination, likeVal);
+        const r = validateEdge(this.plugin, source, destination, likeVal);
+
+        if (cleanup) {
+            this.core.environment.delete(likeVal)
+        }
+
+        return r;
     }
     constructor(public core: Model, public readonly plugin: Plugin) {
         this.activeEdgeType = plugin.types.edges.types.values().next().value;
--- a/app/models/graph-controller.ts
+++ b/app/models/graph-controller.ts
@@ -81,10 +81,10 @@ export class GraphController {
         }
 
         // TODO: ...like
+        const likeVal = new ElementValue(this.activeEdgeType, this.core.environment);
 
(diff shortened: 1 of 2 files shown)


app/models/graph-controller.ts
```diff
@@ -81,10 +81,10 @@ export class GraphController {
         }
 
         // TODO: ...like
+        const likeVal = new ElementValue(this.activeEdgeType, this.core.environment);
 
-        return validateEdge(this.plugin, source, destination, undefined);
+        return validateEdge(this.plugin, source, destination, likeVal);
```
ilargierdi: I'm pretty sure you want to use the type of the provided `like` parameter instead of the active edge type. The purpose of `like` is for the case of moving an edge, which is not necessarily the active edge type.

slaymaker1907: I've added an addition check if a bridge exists and only use the active edge type if one doesn't exist in my latest commit.

slaymaker1907:
Going to just merge this if no one reviews by Sunday.

app/models/graph-controller.ts
```diff
@@ -81,10 +81,20 @@ export class GraphController {
...
+            const bridge = this.bridges.getB(like);
+            if (bridge) {
+                likeVal = bridge.core;
+            }
+        }
 
-        return validateEdge(this.plugin, source, destination, undefined);
-    }
+        if (!likeVal) {
+            likeVal = new ElementValue(this.activeEdgeType, this.core.environment);
```
RighteousRaichu: This is concerning because once we're done with ElementValue's we need to explically remove them from the environment. The like parameter should probably be just ElementType | ElementValue | undefined ?

app/models/graph-controller.ts
```diff
@@ -81,10 +81,20 @@ export class GraphController {
         }
 
         // TODO: ...like
```
RighteousRaichu: We should remove the TODO

app/models/graph-controller.ts
ilargierdi: In that regard, is it possible to get the type information from the arguments of `validateEdgeHandler`? If so, maybe rename `validateEdge` to `typecheckEdge`, only pass element types to `typecheckEdge` and remove the plugin parameter, then `return typecheckEdge(...) && (plugin.validateEdge ? plugin.validateEdge(...) : true)` in `validateEdgeHandler`.

I think this still does not address your concern, though. If `like` is provided as a previously deleted edge, is there any way to get its type information from the environment? I'm not sure this is currently possible.

Sheyne:
We don't need to worry about the CoreElement being deleted as long as there
is a reference to it. It may be removed from the environment though.

app/models/graph-controller.ts
slaymaker1907: @Sheyne said that we should keep it as the edge since the edge could contain more specific information.

PR #364, opened 2017-05-09
Linux package manager

ilargierdi:
Fixes issues with package management on Linux (and potentially other OS's).

* copying to the same source and destination is now skipped
* subdirectories without a `package.json` are no longer attempted to be imported
* removed file filter from `requestOpenDirs`
* cleaned `somePromises` a little
* removed uglifyjs from production build (see inline comment)

### DIFF
--- a/app/services/plugin.service.ts
+++ b/app/services/plugin.service.ts
@@ -12,7 +12,7 @@ import { PythonPluginLoader } from "sinap-python-loader";
 const app = remote.app;
 const LOG = getLogger("plugin.service");
 
-export const PLUGIN_DIRECTORY = IS_PRODUCTION ? path.join(app.getPath("userData"), "plugins") : "./plugins";
+export const PLUGIN_DIRECTORY = IS_PRODUCTION ? path.join(app.getPath("userData"), "plugins") : path.join(app.getAppPath(), "plugins");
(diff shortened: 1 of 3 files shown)


app/services/plugin.service.ts
```diff
@@ -254,8 +257,13 @@ export class PluginService {
     public async importPlugin(dir: string): Promise<void> {
         // Recursively progress through directories until we get interpreter info.
         const dest = path.join(PLUGIN_DIRECTORY, path.basename(dir));
-        LOG.log(`Importing plugins from ${dir} to ${dest}.`);
-        await copy(dir, dest);
+        if (dest === dir) {
+            return new Promise<void>((resolve, reject) => resolve());
```
slaymaker1907: Just return Promise.resolve()

webpack.config.js
```diff
@@ -60,14 +60,19 @@ module.exports = (env = {}) => { // pass command line arguments like `webpack ..
         plugins: [
             new webpack.NoEmitOnErrorsPlugin(),
 
+            /*** # 5/9/2017                                                                     ***/
```
slaymaker1907: I think we really don't even need Uglify for our application since our code is so minimal compared to the rest of electron.

PR #368, opened 2017-09-12
Use new typescript loader

slaymaker1907:
Also fixed sass loader to support Node version 8.


slaymaker1907:
I'm going to merge this over the weekend if no one reviews it by then @Sheyne @danielscottjames @cj-dimaano

PR #370, opened 2021-03-21
Make Sinap run as a web app

danielscottjames:
This will probably never really be complete but I figured I'd push it so I don't eventually lose it. This makes Sinap work as a web app, but the plugins are hardcoded in. I have it on my server at https://sinap.danieljames.dev to play around with. 

It uses the relatively new File System Access API which I've only tested with recent versions of Chrome.

See the branch with the same name I pushed in sinap-typescript-loader which removes Node specific capabilities if you are ever interested in building this yourself locally.


slaymaker1907:
Cool idea, I've been meaning to play around with that API myself. Is this something that is able to be merged or is this PR just for reference to that branch?

danielscottjames:
This branch has all plugin management stripped and they’re just included in the main bundle.
I did this a year or so ago and just wanted to push it and show ya’ll, but I don’t have any plans to finish it rn

